PR #2, opened 2020-05-10
user functional indexes on username and email

hernanat:
This commit updates the indexes on `username` and
  `email` on the `users` table to be functional indexes
  that user the `LOWER` function to allow for case-insensitive
  exact matches.

We could have used `citext` for this, but that is postgres-specific,
  and there is desire amongst the community to be able to port the
  application to other DBMSs in the future without huge issues (I
  realize that it will still be a pain in the ass, but I'll do what I
  can to support this regardless).


hernanat:
[Functionally unique indexes on username / email](https://trello.com/c/E0MsTk40/22-functionally-unique-indexes-on-username-email)

PR #3, opened 2020-05-10
don't add text datatype by default

hernanat:
Since I almost exclusively work with postgres,
  I have a habit of defaulting to the text datatype
  given the way that it is treating by the dbms (I
  won't get into that here).

Since we're aiming to not tie ourselves to a specific
  db implementation, I'm changing this to support that
  goal.

The `text` datatype, then, should only be used for text
  strings that are sufficiently large (i.e. not emails
  and usernames).


hernanat:
[don't default to text datatypes](https://trello.com/c/uONrhWag/23-dont-default-to-text-datatypes)

PR #4, opened 2020-05-10
adds basic submissions model

hernanat:


hernanat:
[As a user I can make a submission](https://trello.com/c/mptWE0bW/2-as-a-user-i-can-make-a-submission)

PR #7, opened 2020-05-12
adds URL Parser

hernanat:
this commit adds a URL parser which is to be used as
  part of the submission creation process.

essentially it's a wrapper around `URI.parse`, with some
  additional functionality.

It:

- enforces the validity of domains (scheme must be present and
  either http or https)

- removes `www\d*` prefixes from domains to prevent multiple entries
  for what amounts to the same domain (from a practical perspective)

- removes common ports for http / https in order to prevent multiple
  entries for the same domain, but with a port included that does not
  make a meaningful distinction (i.e. http + port 80)

- removes (some) known tracking query parameters. We'll obviously need
  to iterate on this

### DIFF
--- a/app/util/url_parser.rb
+++ b/app/util/url_parser.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+class UrlParser
+  def self.parse(url)
+    new(url).parse if url.present?
+  end
+
+  def initialize(raw_url)
+    @raw_url = raw_url
+  end
+
+  attr_reader :raw_url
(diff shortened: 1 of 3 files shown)


hernanat:
[URL Parser for submissions](https://trello.com/c/NeuxNcaU/24-url-parser-for-submissions)

hernanat:
[URL Parser for submissions](https://trello.com/c/NeuxNcaU/24-url-parser-for-submissions)

PR #8, opened 2020-05-12
adds Domain model

hernanat:
This commit adds a model to represent the underlying
  `domain` that a submission comes from.

This is useful for being able to do things like ban URLs
  from a given domain (if it's found to have malware, creepier
  than usual tracking, bad informatin, etc.).

It also adds a foreign key reference to the submissions table.

Submissions may or may not have a domain, depending on whether or
  not it is a URL or text submission.

### DIFF
--- a/app/models/domain.rb
+++ b/app/models/domain.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Domain < ApplicationRecord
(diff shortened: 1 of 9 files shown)


hernanat:
[add Domain model for tracking domains for submitted urls](https://trello.com/c/ZfXdrHra/25-add-domain-model-for-tracking-domains-for-submitted-urls)

PR #9, opened 2020-05-13
adds URL Health Check Service

hernanat:
This commits adds in a (naive) URL health check service.

Basically, when the user submits a URL, we want to make sure that
  it's valid / the destination exists, so we try to make a request.

In the case of redirects, the service also keeps track of the final
  destination so that we can store it directly instead of forcing
  users to follow all the redirects.

We also disallow the use of IPs, because IPs are not necessarily
  long lived like URLs (someone can move their application to another
  server and keep the same domain, the same isn't true for an IP).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,9 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby '2.7.1'
 
+# http
+gem "faraday"
+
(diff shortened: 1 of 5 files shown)


hernanat:
[URL Healthcheck service](https://trello.com/c/pSV0hfwQ/26-url-healthcheck-service)

PR #11, opened 2020-05-14
Potential style idea

ashpil:
Hey,

Here's an idea for the design of the site. 

image

I was trying to think of what "math-inspired design would be", and one idea that came up was a LaTeX style since that's basically universal for math. So I tried to come up with a LaTeX-inspired design style that looked like an intentional design choice rather than a lack of one. I'm not set in stone on this, but I think it looks okay and clean and am wondering what you think.

Also, if you do like the idea, I could not figure out how to use a custom font with @font-face and Rails (so to design this I just used a local change on my machine). I understand you may object to loading a custom font as it's somewhat unnecessary styling, but think of it this way: you're already loading this same font for LaTeX support. Ideally we'd figure out a way for it to not load the font twice and just used the same one for built-in LaTeX and normal text.

Couple of things I'm unsure about: 

1. Colors. LaTeX doesn't really use colors but websites do, so I'm unsure what the ideal palette would be.

2. I'm not a big fan of the large boxy border on focus. I understand it's very good for keyboard navigators, but it doesn't seem ideal for someone who typically uses a mouse. I think there's some way to display the large box for keyboard-navigation only, but I haven't looked into the details.

Let me know what you think, @hernanat !

Thanks,
ashpil/brunch

### DIFF
--- a/app/assets/stylesheets/base/_layout.scss
+++ b/app/assets/stylesheets/base/_layout.scss
@@ -17,13 +17,15 @@ body {
 
 .application-container {
   @include padding(var(--spacing--small));
-  width: 50%;
+  width: 70%;
   margin: auto;
+  max-width: 80rem;
 }
 
 @media only screen and (max-width: 1155px) {
   .application-container {
     margin-top: 175px;
-    width: 60%;
+    max-width: 95%;
+    width: 100%;
   }
 }
--- a/app/assets/stylesheets/base/_variables.scss
+++ b/app/assets/stylesheets/base/_variables.scss
@@ -1,11 +1,13 @@
 // Colors
 $blue: #1565c0;
 $blue-light: #3b83d5;
+$blue-light-very: #f4fbff;
 $congress-blue: #004e98;
 $gray: #333;
 $off-indigo: #4f86c6;
 $scorpion: #606060;
 $white: #fff;
+$white-gray: #fafafa;
 
 // Font Colors
 $font-color--base: $gray;
@@ -14,15 +16,16 @@ $action-color--alt: $blue-light;
 $action-color--contrast: $white;
 
 // Background Colors
-$base-background-color: #EBEBEB;
+$base-background-color: white;
 $viewport-background-color: $base-background-color;
 
 // Primary color (nav, etc.)
 $primary-color: $congress-blue;
 
 :root {
+
   // Typography
-  --font-family: system-ui, sans-serif;
+  --font-family: "CMU Serif", system-ui, sans-serif;
   --font-family--heading: var(--font-family);
 
   --font-weight--thin: 100;
@@ -47,7 +50,7 @@ $primary-color: $congress-blue;
 
   // Borders
   --border-width: 0.0625rem;
-  --border: var(--border-width) solid currentColor;
+  --border: var(--border-width) solid #333;
 
   // Focus
   --focus-outline-width: 0.1875rem;
--- a/app/assets/stylesheets/base/_buttons.scss
+++ b/app/assets/stylesheets/base/_buttons.scss
@@ -21,6 +21,8 @@ button,
   user-select: none;
   vertical-align: middle;
   white-space: nowrap;
+  font-variant: small-caps;
+  margin: 0.5rem 0;
 
   &:hover {
     background-color: $action-color--alt;
--- a/app/assets/stylesheets/base/_forms.scss
+++ b/app/assets/stylesheets/base/_forms.scss
@@ -1,8 +1,3 @@
-:root {
-  --form-box-shadow: inset 0 --border-width 0.1875rem rgba(#000, 0.06);
-  --form-box-shadow-focus: var(--form-box-shadow), 0 0 0.3125rem #{$action-color};
-}
-
 fieldset {
   background-color: transparent;
   border: 0;
@@ -47,10 +42,9 @@ textarea {
 input:not([type]),
 textarea {
   appearance: none;
-  background-color: transparent;
+  background-color: $white-gray;
   border: var(--border);
   border-radius: var(--border-radius);
-  box-shadow: var(--form-box-shadow);
   box-sizing: border-box;
   margin-bottom: var(--spacing--small);
   padding: calc(var(--spacing) / 3);
@@ -58,7 +52,7 @@ textarea {
   width: 100%;
 
   &:focus {
-    box-shadow: var(--form-box-shadow-focus);
+    background-color: $blue-light-very;
   }
 
   &:disabled {
@@ -85,7 +79,7 @@ textarea {
 
 [type="checkbox"],
 [type="radio"] {
-  display: inline;
+  display: inline-block;
   margin-right: var(--spacing--small);
 }
 
--- a/app/assets/stylesheets/pages/_submissions.scss
+++ b/app/assets/stylesheets/pages/_submissions.scss
@@ -9,7 +9,6 @@
           text-decoration: none;
           font-size: 1.2rem;
           font-weight: var(--font-weight--bold);
-          font-color: $primary-color;
         }
       }
       .submission-line2 {
--- a/app/assets/stylesheets/shared/_navbar.scss
+++ b/app/assets/stylesheets/shared/_navbar.scss
@@ -1,13 +1,14 @@
 // for the love of everything that is holy someone please make this nicer
 .navbar {
   @include padding(null var(--spacing--small) null var(--spacing--small));
-  background-color: $primary-color;
-  height: 55px;
-  line-height: 55px;
+  height: 60px;
+  line-height: 60px;
+  box-shadow: 0px 5px 5px -4px rgba(0,0,0,0.2);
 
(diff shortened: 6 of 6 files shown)


hernanat:
Hey @ashpil, this looks great. I’d for sure be down with this design. In terms of the color scheme I think this is fine, and we could iterate on it if we want. I like the simplicity. Idk if you’ve run the rails app locally recently and seen what Ive done so far.

We can have custom font assets no problem at all. Rails 6 uses webpacker so we can configure it there, or at it under the public directory. It may be better to keep it with webpacker. If you’re struggling to do that I can help you out on a call / give you some more info via email, or you can try it out again.

With respect to boxy borders, are you referring to the input boxes? I’m sure we can find a middle ground there. It does help in terms of accessibility to have things like border highlighting, but we also don’t want it to be obnoxious.

hernanat:
Ill have a look at the code in a bit, currently on my phone!

ashpil:
If you could just link what the proper modern way for Rails 6 to include custom fonts is, I'm sure I could handle it. I'm not sure how the current LaTeX implementation works, but ideally we'd just link that same font file, as we don't want to load two of the same thing. That might be a bit more complicated than just the typical custom font asset implementation. I'm not sure if I could figure that out.

By boxy focus border I mean this:

image

It's good for tabbing through and keyboard navigation, but ideally you'd have something more subtle for people who are just clicking on stuff with a mouse, otherwise it is sort of distracting. 

Also, quick question, how are you choosing when to use SASS variables and when to use normal CSS ones? Because it seems to be a mix of both right now, so I'm wondering what the choices behind that are.

hernanat:
> If you could just link what the proper modern way for Rails 6 to include custom fonts is, I'm sure I could handle it.

[This guide](https://github.com/rails/webpacker/blob/master/docs/assets.md) should have all of the information that you need. If your fonts aren't part of a package I believe you have to stuff them under either `app/assets/fonts` or `app/javascript/fonts` (this makes me feel obligated to point out that I think the `app/javascript` folder for webpacker related things is a mistake because it's used for way more than that these days). Otherwise you would just use the non-relative `~` syntax described in the guide.

Alternatively, if you put them all under `public/fonts` we can skip the usage of webpacker altogether. I'm not sure it matters one way or another in this case.

> I'm not sure how the current LaTeX implementation works, but ideally we'd just link that same font file, as we don't want to load two of the same thing.

It's true that the [katex](https://katex.org/) package ships with fonts, although I think each type has a specific purpose (there are a good number of them):
Screen Shot 2020-05-14 at 10 19 22 AM

> By boxy focus border I mean this:

Oh...yeah...I see what you mean. I'm not really sure what to do in this case. On the one hand, YUCK. On the other hand, I really would like to make this site as [accessible](https://usability.yale.edu/) as possible, and [focus style for links](https://usability.yale.edu/web-accessibility/articles/links#focus) is an important part of this (for reasons identical to as you mentioned -- keyboard users).

> It's good for tabbing through and keyboard navigation, but ideally you'd have something more subtle for people who are just clicking on stuff with a mouse, otherwise it is sort of distracting.

I believe [focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) is what we want in this case.

hernanat:
> Also, quick question, how are you choosing when to use SASS variables and when to use normal CSS ones? Because it seems to be a mix of both right now, so I'm wondering what the choices behind that are.

This is laziness on my part, and I apologized. I used [bitters](https://github.com/thoughtbot/bitters) to get some of the styles up and running. And for whatever reason some of the generated variables were CSS variables. Ideally I would like to use sass variables throughout the codebase as they are much nicer. If this is something that you would like to change as part of what you're doing, feel free. Otherwise I will open up a PR to clean up my own mess :)

app/assets/stylesheets/base/_layout.scss
```diff
@@ -17,13 +17,15 @@ body {
...
+  width: 70%;
   margin: auto;
+  max-width: 80rem;
 }
 
 @media only screen and (max-width: 1155px) {
   .application-container {
     margin-top: 175px;
-    width: 60%;
+    max-width: 95%;
```
hernanat, line 28: do `width` and `max-width` actually play together? I was always under the impression that only one could be used at a time (I'm not much of a css expert).

Also, could you attach a screen shot to this PR of what this translates to on smaller screen views? And maybe as a sanity check just make sure that everything sizes up and down as expected if you haven't already (in particular with the navbar + submissions on the page -- when I was first working on the layout the navbar came dow over the submissions on small screens, oops 😇 )

app/assets/stylesheets/base/_variables.scss
```diff
@@ -47,7 +50,7 @@ $primary-color: $congress-blue;
 
   // Borders
   --border-width: 0.0625rem;
-  --border: var(--border-width) solid currentColor;
+  --border: var(--border-width) solid #333;
```
hernanat, line 53: what do you think about using the defined `$gray` variable here since we have it defined above?

hernanat:
also, I'm going to add you as a contributor to the repo so you don't have to keep a forked codebase. I think that would make this much easier.

app/assets/stylesheets/base/_layout.scss
ashpil: They do play together, as I understand it. The mobile case doesn't actually have a reason to be using this set up, oops, but for non-mobile what it does is it makes the main content 70% of the screen, but if the screen is too large, it doesn't go beyond 80rem and keeps it as a centered column of 80rem in the center. I didn't bother too much with optimizing the mobile set up, as I figured the end product would be having the menu as a drop down or something anyway.

app/assets/stylesheets/base/_variables.scss
ashpil: Actually, this works basically the same with currentColor and I forgot to change it back. 

For some reason the '$gray' variable doesn't seem to work well with CSS variables - it's transferred into the form as just '0.0625rem solid $gray' rather than '0.0625rem solid #333'. This is why my question of why you're sometimes using Sass variables and sometimes using CSS ones came up, as they don't seem to mesh too well together.

hernanat: ahh yeah. in this case we'd also have to change this border variable to a sass variable as well. I won't ask you to do all that here, I'll open a PR with those changes once yours are in (unless you feel like making those changes in which case feel free)

app/assets/stylesheets/base/_layout.scss
hernanat: > as I figured the end product would be having the menu as a drop down or something anyway.

I'm on the fence about a dropdown / hamburger button. On the one hand, the pattern is (for better or for worse; I believe for worse) ubiquitous for small-screen styles of applications, and so well recognized / understood. On the other hand, it requires extra clicks from users in order to navigate between pages, and otherwise hides navigation information. Accessibility also comes into play here -- typical hamburger implementations are not accessible (not that people do this on purpose), so it would require some additional work to support that.

That being said I don't want to dictate whether or not we have a dropdown / hamburger or something like what currently exists (it just makes the navbar extend over several lines for smaller screens). What are your thoughts on the matter?

ashpil: Honestly I'm not set on the drop down button either. I agree that it's not ideal. I'm just not sure what a better alternative is. I suppose an option is just making the text smaller and trying to fit it into one row, because it does take up a lot of space now. I'm surprised that  a drop down button is not accessible, though, as as you said, it's so ubiquitous. 

I'll play around a bit with the design and see what I can come up with.

ashpil:
I'm wondering why you haven't used flexbox at all? I guess there hasn't really been a particularly strong need for it anywhere, but I typically use a lot of flexbox/css-grid in my designs but I've avoided adding any of those here because I didn't see it in the existing code. Are you specifically avoiding those or have you just not felt a need to use it?

app/assets/stylesheets/base/_layout.scss
hernanat: > I'm surprised that a drop down button is not accessible, though, as as you said, it's so ubiquitous.

I should emphasize that the usual implementations of these hamburger style menus are not accessible by default. We would have to make sure that ours was.

Not sure if you've seen what I've did for small screens as of now, but basically I just have the navbar extend over several lines. In other words, everything is still visible, it just gets its own line. This turns out relatively fine in our case since we don't have a great number of links on our navbar anywys.

hernanat:
> I'm wondering why you haven't used flexbox at all? I guess there hasn't really been a particularly strong need for it anywhere, but I typically use a lot of flexbox/css-grid in my designs but I've avoided adding any of those here because I didn't see it in the existing code. Are you specifically avoiding those or have you just not felt a need to use it?

Neither of these, actually. A lot of the work that I've done has used float-based grid styles, and I just haven't gotten around to getting comfortable with flexbox / css-grid like I should have.

If you think that their usage would improve the quality of the code base / be better from the perspective of modernity, I am fully on board with that. I should probably be using them anyways.

PR #13, opened 2020-05-14
Adds the ability to submit and view submissions

hernanat:
This commit adds the ability for users to create submissions
  as well as view existing submissions.

It includes (naive) support for both markdown and LaTeX.

The way we process the submission bodies for display should
  probably be thoroughly QA'd to make sure we're not doing
  anything insane.

I opted to format the markdown at render time due to the fact
  that certain markdown elements requre a large number of
  HTML tags to render, and this incurrs an additional storage
  cost for the submission body. If this becomes an issue in the
  future we can look into caching the markdown html somewhere.

### DIFF
--- a/app/assets/stylesheets/base/_forms.scss
+++ b/app/assets/stylesheets/base/_forms.scss
@@ -104,3 +104,12 @@ select {
     outline: none;
   }
 }
+
+.errors {
+  list-style-type: circle;
+  margin-left: $spacing;
+  margin-bottom: $spacing-smaller;
+  .error {
+    color: $red;
+  }
+}
--- a/app/assets/stylesheets/pages/_submissions.scss
+++ b/app/assets/stylesheets/pages/_submissions.scss
@@ -1,40 +1,60 @@
 .submissions-list {
   @include padding($spacing-smaller null null null);
+}
 
-  ol {
-    li {
-      @include padding(null null $spacing-small null);
-      .submission-line1 {
-        .submission-link a {
-          text-decoration: none;
-          font-size: 1.2rem;
-          font-weight: $font-weight-bold;
-        }
-      }
-      .submission-line2 {
-        .submission-info, .submission-comments {
-          color: $scorpion;
-          a {
-            text-decoration: none;
-            color: $scorpion;
-          }
-        }
+.submission {
+  @include padding(null null $spacing-small null);
+  .submission-info, .submission-comments {
+    color: $scorpion;
+    a {
+      text-decoration: none;
+      color: $scorpion;
+    }
+  }
+  .submission-line1 {
+    .submission-link a {
+      text-decoration: none;
+      font-size: 1.2rem;
+      font-weight: $font-weight-bold;
+      margin-right: $spacing-xs;
+    }
+  }
+  .submission-line2 {
+    .submission-time::after {
+      content: "|";
+    }
 
-        .submission-time::after {
-          content: "|";
-        }
+    .submission-info a::after {
+      @include margin(null 1px null null);
+      @include padding(null null null 4px);
+      content: "|";
+      font-size: 1rem;
+    }
+    .submission-user {
+      color: $off-indigo;
+      a { text-decoration: none; }
+    }
+  }
+}
 
-        .submission-info a::after {
-          @include margin(null 1px null null);
-          @include padding(null null null 4px);
-          content: "|";
-          font-size: 1rem;
-        }
-        .submission-user {
-          color: $off-indigo;
-          a { text-decoration: none; }
-        }
-      }
+.submission-body {
+  border-top: $light-border;
+  padding-top: $spacing-smaller
+}
+
+.original-author-statement {
+  padding-top: $spacing-smaller;
+  margin-bottom: $spacing-smaller;
+  .left-hand-label {
+    display: inline-block;
+  }
+}
+
+@media only screen and (max-width: 1155px) {
+  .submissions-list {
+    .submission {
+      border-bottom: $light-border;
+      margin-bottom: $spacing-smaller;
     }
   }
 }
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -1,9 +1,32 @@
 class SubmissionsController < ApplicationController
+  before_action :authenticate_user!, only: [:new, :create]
+
   def index
     @paginator = SubmissionSearch.new(submission_search_params).results_paginator
   end
 
+  def show
+    @submission = Submission.preload(:user, :domain).find(params[:id])
+  end
+
+  def new
+    @form = CreateSubmissionForm.new({}, current_user)
+  end
+
+  def create
+    @form = CreateSubmissionForm.new(create_submission_params, current_user)
+    if @form.save
+      redirect_to submission_path(@form.submission.id)
+    else
+      render :new
+    end
+  end
+
   private
+
+  def create_submission_params
+    params.require(:create_submission_form).permit(:url, :title, :body, :original_author)
+  end
 
   def submission_search_params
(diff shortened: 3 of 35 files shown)


app/assets/stylesheets/base/_forms.scss
```diff
@@ -108,3 +108,12 @@ select {
     outline-offset: var(--focus-outline-offset);
   }
 }
+
+.errors {
```
hernanat, line 112: @ashpil I'd love to get your feedback on some of this CSS stuff. Let me know if you're able, otherwise we could always revisit later.

app/form_objects/create_submission_form.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
+
+# this should probably be revisted sometime in the
+# near future
+class CreateSubmissionForm
```
hernanat, line 5: this is sort of a mess right now, I'll clean it up at some point in the future. In the meantime I've tested it pretty thoroughly so it should work fine (famous last words).

app/javascript/packs/latex.js
```diff
@@ -0,0 +1,244 @@
...
+
+    // Enable sharing of global macros defined via `\gdef` between different
+    // math elements within a single call to `renderMathInElement`.
+    optionsCopy.macros = optionsCopy.macros || {};
+
+    renderElem(elem, optionsCopy);
+};
+
+const callback = function() {
+  const elements = document.getElementsByClassName("latex-container");
```
hernanat, line 230: I'm being deliberate about looking for specific parts of the DOM to try to render latex in. For one, so that we don't unintentionally render latex in areas where it shouldn't be, and two, so that we don't have to parse the entire DOM.

app/views/submissions/show.html.haml
```diff
@@ -0,0 +1,5 @@
+= render "submission_list_item", submission: @submission
+.submission-body.latex-container
+  = sanitize CommonMarker.render_html(@submission.body.gsub("\\", "\\\\\\")) if @submission.body.present?
```
hernanat, line 3: this is incredibly naive and nasty I know. We can come back to it. Without this, folks would have to double escape their latex commands, which I don't think is a very good user experience.

hernanat:
Here's a demo of markdown / latex rendering in action:

Screen Shot 2020-05-14 at 4 04 42 PM

Screen Shot 2020-05-14 at 4 04 54 PM

hernanat:
I just realized that the markdown lists aren't rendering properly because we currently have styles overriding the way lists are displayed 🤦

app/assets/stylesheets/base/_forms.scss
ashpil: This looks good, except that in my PR I changed all CSS variables to Sass ones, and you're relying quite heavily on CSS variables.

hernanat: yeah, let's get your changes in first and then I'll update my PR to fall in line with them

ashpil:
> 
> 
> Here's a demo of markdown / latex rendering in action:
> 
> Screen Shot 2020-05-14 at 4 04 42 PM
> 
> Screen Shot 2020-05-14 at 4 04 54 PM

By the way, how do you create a user account to test that stuff?

hernanat:
you can either do it manually in the rails console:

```ruby
u = User.create(email: "foo@bar.com", password: "abcd12345", password_confirmation: "abcd12345")
u.confirm
```

or you can go to `http://localhost:3000/users/sign_up` (or just click on the `Login Button` and click the sign up link). we don't currently have registrations disabled (or a mechanism for disabling / enabling them on the fly)

ashpil:
I tried the second way but I wasn't able to get past the email confirmation part. I'll do the console way, seems simpler.

hernanat:
right, sorry. You'd then have to go into the console and manually confirm the user.

ashpil:
Okay, so update this to use the new Sass variables and we can probably merge it?

hernanat:
Yep will do in a bit, not near my computer right now.

hernanat:
Screen Shot 2020-05-14 at 9 21 56 PM

@ashpil I had the idea of putting borders around the submissions on the index for smaller screens, but now I'm not so sure. What do you think?

app/assets/stylesheets/pages/_submissions.scss
```diff
@@ -1,40 +1,60 @@
...
-      }
-      .submission-line2 {
-        .submission-info, .submission-comments {
-          color: $scorpion;
-          a {
-            text-decoration: none;
-            color: $scorpion;
-          }
-        }
+.submission {
```
hernanat, line 5: I originally had the styling scoped to an `li`, but we're using the same styling on the listing page itself which doesn't exist within the context of a list.

hernanat:
Screen Shot 2020-05-14 at 9 26 58 PM

looks awesome with your styling btw

app/assets/stylesheets/pages/_submissions.scss
```diff
@@ -1,40 +1,60 @@
...
+  }
+  .submission-line1 {
+    .submission-link a {
+      text-decoration: none;
+      font-size: 1.2rem;
+      font-weight: $font-weight-bold;
+      margin-right: $spacing-xs;
+    }
+  }
+  .submission-line2 {
```
hernanat, line 22: does the styling for these make sense? i.e. is there a more desirable approach to style these line items?

app/controllers/submissions_controller.rb
```diff
@@ -1,10 +1,33 @@
 class SubmissionsController < ApplicationController
+  before_action :authenticate_user!, only: [:new, :create]
```
hernanat, line 2: I haven't given the ability to edit yet. Not sure if we want to, and what / for how long a submission should be editable after submitting.

ashpil:
> 
> 
> Screen Shot 2020-05-14 at 9 21 56 PM
> 
> @ashpil I had the idea of putting borders around the submissions on the index for smaller screens, but now I'm not so sure. What do you think?

I like this. I like that there's some subdivision of content. Not sure if this is the perfect way to do it, but better than none, imo. And as you said, we can always iterate.

ashpil:
> I like this. I like that there's some subdivision of content. Not sure if this is the perfect way to do it, > but better than none, imo. And as you said, we can always iterate.

But of course it may end up looking different if we add thumbnails and tags and more info.

hernanat:
Im on the fence on whether or not to do tags before comments

app/controllers/submissions_controller.rb
ashpil: I think the ability to edit would be nice. Just from a perspective of users being able to clarify things (or maybe if we wanted that we could allow the user to do a sort of pinned comment for their own post like a mod can do on Reddit?).

Another option is for edit history to be visible.

hernanat:
If you think this is good to go lmk and Ill start working on what’s next!

ashpil:
Sorry, I was just getting it on my machine and checking it out! Looks good! I'll spend some time tomorrow doing the CSS for the Markdown, unless you're interested in doing it?

hernanat:
No problem, sorry about that. Thanks for the review / approval.

You can take care of that if you have the bandwidth, I think I want to take care of tags real quick and then move onto the comments :D

hernanat:
[As a user I can make a submission](https://trello.com/c/mptWE0bW/2-as-a-user-i-can-make-a-submission)

hernanat:
[As a user I can view submissions](https://trello.com/c/PoECQzJc/19-as-a-user-i-can-view-submissions)

ashpil:
[As a user I can format a submission / comment using LaTeX](https://trello.com/c/EBpIRrJ8/9-as-a-user-i-can-format-a-submission-comment-using-latex)

PR #14, opened 2020-05-16
adds ability for users to tag submissions

hernanat:
https://trello.com/c/mvKg5bLf

this commit allows users to select up to 5 tags
  for their submissions.

other rules:

- minimum of 1 tag must be selected

- minimum of 1 non-media (i.e. filetype) tag must be
  selected

- non-admin / mods are restricted from accessing mod tags

A follow up PR will make the tag links clickable, which
  will take the user to the submission index view, but only
  showing submissions with the clicked-on tag.

### DIFF
--- a/app/assets/stylesheets/shared/_tags.scss
+++ b/app/assets/stylesheets/shared/_tags.scss
@@ -1,0 +1,43 @@
+.tag {
+  text-decoration: none;
+  border-radius: 0.25rem;
+  padding: 0.15rem;
+  font-size: 65%;
+  vertical-align: middle;
+}
+
+@mixin tag($color) {
+  background-color: $color;
+  border: 1px solid darken($color, 40%);
+  color: darken($color, 60%);
+  transition: color $transition;
+  transition: background-color $transition;
+  &:hover {
+    background-color: lighten($color, 10%);
+    color: darken($color, 50%);
+  }
+  &:active {
+    background-color: darken($color, 10%);
+    color: darken($color, 80%);
+  }
+}
+
+.tag_topic {
+  @include tag(#ceecfd);
+}
+
+.tag_media {
+  @include tag(#d9c7f2);
+}
+
+.tag_source {
+  @include tag(#f7f5a8);
+}
+
+.tag_meta {
+  @include tag(#c7f2cd);
+}
+
+.tag_mod {
+  @include tag(#ff8f8f);
+}
--- a/db/migrate/20200515180402_create_tags.rb
+++ b/db/migrate/20200515180402_create_tags.rb
@@ -1,0 +1,10 @@
+class CreateTags < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tags, id: :string do |t|
+      t.string :description, limit: 100
+      t.integer :kind, null: false, default: 0
+
+      t.timestamps
+    end
+  end
+end
--- a/db/migrate/20200515182627_create_submission_tags.rb
+++ b/db/migrate/20200515182627_create_submission_tags.rb
@@ -1,0 +1,10 @@
+class CreateSubmissionTags < ActiveRecord::Migration[6.0]
+  def change
+    create_table :submission_tags do |t|
+      t.belongs_to :submission, null: false, foreign_key: { on_delete: :cascade }
+      t.belongs_to :tag, null: false, type: :text, foreign_key: { on_delete: :cascade }
+
+      t.timestamps
+    end
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -64,6 +64,9 @@ Naming/PredicateName:
 Naming/RescuedExceptionsVariableName:
   Enabled: false
 
+Naming/MemoizedInstanceVariableName:
+  Enabled: false
+
 Naming/MethodParameterName:
   Enabled: false
 
--- a/app/assets/stylesheets/pages/_submissions.scss
+++ b/app/assets/stylesheets/pages/_submissions.scss
@@ -13,11 +13,18 @@
     }
   }
   .submission-line1 {
+    position: relative;
     .submission-link a {
       text-decoration: none;
       font-size: 1.2rem;
       font-weight: $font-weight-bold;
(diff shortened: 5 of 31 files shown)


hernanat:
[As a user I can tag a submission when I create it](https://trello.com/c/mvKg5bLf/11-as-a-user-i-can-tag-a-submission-when-i-create-it)

hernanat:
I know this is fugly right now. I would like to have a stylish tag select input box where users can type tags and the available ones would show up, but I am not a fan of some of the existing libraries for doing this (I played around with some earlier and they're not very accessible, and the one that is accessible is a single select by default that would require me to write a multiselect "engine" which at that rate I'll just write my own widget, which I plan on doing). It's worth noting that there are a metric crap ton of jquery widgets to do this, but I'm specifically trying to avoid introducing jquery into this project.

Screen Shot 2020-05-15 at 8 44 52 PM

Screen Shot 2020-05-15 at 8 45 12 PM

app/assets/stylesheets/shared/_tags.scss
```diff
@@ -0,0 +1,34 @@
+.tag {
+  text-decoration: none;
+  border-radius: 0.25rem;
+  padding: 0.15rem;
+  font-size: 65%;
+  vertical-align: middle;
+}
+
+.tag_topic {
```
hernanat, line 9: @ashpil don't hate me for my poor color choices 🙈

hernanat:
the screenshot I attached shows the types of tags from left of right as follows:

- topic
- media
- source
- meta
- mod

db/migrate/20200515182627_create_submission_tags.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSubmissionTags < ActiveRecord::Migration[6.0]
+  def change
+    create_table :submission_tags do |t|
```
hernanat, line 3: I thought about caching tag ids / info directly on submissions and using a gin index to avoid having to create a join table, but that seemed like a premature optimization, and I'm also not sure how that would translate into mysql (since we don't want to make DBMS-specific design choices), and I didn't really feel like trying to figure it out at this point.

db/migrate/20200515180402_create_tags.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTags < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tags, id: :string do |t|
```
hernanat, line 3: there's not much to these. I opted to just have the tag itself be the id / primary key of the table. We also have a description that is used as the title in link elements to give more information on-hover.

ashpil:
> I played around with some earlier and they're not very accessible, and the one that is accessible is a single select by default that would require me to write a multiselect "engine" which at that rate I'll just write my own widget, which I plan on doing

What do you see as an accessible version of that?

app/assets/stylesheets/shared/_tags.scss
ashpil: I'm sure there's a custom that all Rails devs know, but how do I create a tag to test them out?

ashpil: I'm sure there's a custom that all Rails devs know, but how do I create a tag to test them out?

hernanat: Heya. So in the Tag model class you’ll see the different kinds of tags I’ve defined so far. To create one, go in the rails console and for example type:

`Tag.topic.create(id: “algebra”, description: “algebra stuff”)`

Similar for others.

I’ll make a commit that updates our seeds file with some so we can just run `rake db:seed` to populate some in the future

ashpil: Thanks! I realize this isn't something that you usually have to specify. 😛

hernanat: Nah it’s all good!

ashpil: Why don't we include the tag color info in the info of the tag itself rather than the CSS file, so we can create a tag without modifying CSS? And make the non-base colors simply like 20% darker or lighter than the tag info, so we don't need to include 3 hex values for each tag?

hernanat: We don’t have to modify the css for every tag, only if we create new tag _kinds_. All tags of the same kind will share the css — for example above you can create as many topic tags as you want and they will share the same css.

ashpil: That's true.

hernanat:
> What do you see as an accessible version of that?

A few things:

- tabbing in / out like a normal control

- tabbing / arrowing _between tags_ -- I noticed in some of the ones that I was playing with that currently exist that you could tab to the first selected option, but for whatever reason trying to tab after that would select _all_ the selected options.

- be able to remove a selected element from the middle / beginning without using a mouse. there was no obvious way to do this when I was playing around with others. For the most part, you are required to remove _all_ of the ones proceeding the one that you want to remove first if you want to use the backspace key

- probably some other ideas but I'm still in the process of waking up :)

ashpil:
Check this out. Should be a bit easier to add new tags in the future with the mixin, and I made them actually centered (centering things in CSS is so ungodly difficult, I have no idea why).

app/assets/stylesheets/shared/_tags.scss
```diff
@@ -6,29 +6,36 @@
   vertical-align: middle;
 }
 
+@mixin tag($color) {
```
hernanat, line 9: dude this is sick!

hernanat:
ffs @ashpil would you mind pushing your changes again. I forgot to pull them before  pushing -.-

ashpil:
No worries, I think I just did it.

hernanat:
merge conflicts from hell phew I think fixed them all

hernanat:
sorry I know you already approved this but I just realized I had forgotten to add system tests to cover the tags being displayed on the index / show pages

hernanat:
thanks !

PR #16, opened 2020-05-16
adds short id to submissions

hernanat:
https://trello.com/c/fFuwQ1Lb

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -70,15 +70,18 @@ Naming/MethodParameterName:
 Rails/HasManyOrHasOneDependent:
   Enabled: false
 
+Rails/NotNullColumn:
+  Enabled: false
+
+Rails/OutputSafety:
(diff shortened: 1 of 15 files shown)


hernanat:
[add short id support for nicer links](https://trello.com/c/fFuwQ1Lb/29-add-short-id-support-for-nicer-links)

hernanat:
Screen Shot 2020-05-16 at 12 19 31 PM

hernanat:
instead of putting arbitrary integers in the URL we can do this instead, which is much nicer :)

ashpil:
This seems like something we can just merge, right?

hernanat:
I think so -- just didn't know if you wanted to take a look / make any comments before I did. If you're good to go then I'll send it in!

hernanat:
thanks @ashpil !

PR #17, opened 2020-05-16
limit rate of which users can make submissions

hernanat:
https://trello.com/c/G5Kr4B5E

this commit adds logic to prevent users from making
  submissions if their last submission was within 10
  minutes of the current time.

the idea is to help prevent users from spamming submissions,
  which will help a great deal in moderating the site.

we can tweak the allowed time between submissions as necessary.


ashpil:
Seems like a good feature, 10 minutes does seem a bit like a lot to me, though. But I'm okay with this until we see how it works/if people complain.

hernanat:
yeah I'm not really sure what a good sweet spot would be. If it does seem like a lot we can change it to 5 or something if you think that'd be better, or we can revisit it later like you mention.

hernanat:
[rate limit submissions to help prevent spam](https://trello.com/c/G5Kr4B5E/30-rate-limit-submissions-to-help-prevent-spam)

hernanat:
thanks @ashpil !

PR #18, opened 2020-05-17
adds the ability for users to hide / save submissions

hernanat:
https://trello.com/c/iNwsvA8X

this commit introduces "submission actions", which at present
  are just a means of hiding / saving submissions for a user.

when the user hides a submission on the index, we will lower its
  opacity, and on the next page load it will not be included in
  the results at all (vs removing it from the dom immediately).

for hidden / saved submissions, both action links will be changed
  to their active status when an action was created by the user for
  a particular submission, ie:

- "hide" -> "hidden" when a user has hidden a submission

- "save" -> "saved" when a user saved a submission (we also change the
  color of the link text)

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -2,11 +2,11 @@ class SubmissionsController < ApplicationController
   before_action :authenticate_user!, only: [:new, :create]
 
   def index
-    @paginator = SubmissionSearch.new(submission_search_params).results_paginator
+    @paginator = SubmissionSearch.new(submission_search_params, current_user).results_paginator
   end
 
   def show
-    @submission = Submission.friendly.preload(:user, :domain).find(params[:short_id])
+    @submission = submission
   end
 
   def new
@@ -23,6 +23,23 @@ class SubmissionsController < ApplicationController
   end
 
   private
+
+  def submission
+    Submission.friendly.preload(:user, :domain).then { |rel|
+      if current_user.present?
+        rel.
+          left_join_saved_info_for(current_user).
+          left_join_hidden_info_for(current_user).
+          select(%(
+            submissions.*,
+            hidden_actions.id AS hidden_action_id,
+            saved_actions.id AS saved_action_id
+          ).squish)
+      else
+        rel.select("submissions.*, NULL AS hidden_action_id, NULL AS saved_action_id")
+      end
+    }.find(params[:short_id])
+  end
 
   def create_submission_params
     params.require(:create_submission_form).
--- a/app/models/concerns/short_id.rb
+++ b/app/models/concerns/short_id.rb
@@ -13,7 +13,7 @@ module ShortId
     short_id = nil
 
     loop do
-      short_id = SecureRandom.urlsafe_base64(6).gsub(/-|_/, ("a".."z").to_a[rand(26)])
+      short_id = "#{self.class.short_id_prefix}#{SecureRandom.base36(8)}"
       break unless self.class.name.constantize.where(short_id: short_id).exists?
     end
 
--- a/app/models/submission_action.rb
+++ b/app/models/submission_action.rb
@@ -1,0 +1,9 @@
+class SubmissionAction < ApplicationRecord
+  enum kind: {
+    hidden: 0,
+    saved: 1
+  }
+
+  belongs_to :submission, foreign_key: :submission_short_id, primary_key: :short_id
+  belongs_to :user
+end
--- a/app/search_objects/submission_search.rb
+++ b/app/search_objects/submission_search.rb
@@ -1,9 +1,12 @@
+# frozen_string_literal: true
+
 class SubmissionSearch
   DEFAULT_PAGE = 0
   DEFAULT_PER_PAGE = 25
 
-  def initialize(params)
+  def initialize(params, user)
     @params = params
+    @user = user
   end
 
   def results_paginator
@@ -12,10 +15,26 @@ class SubmissionSearch
 
   private
 
-  attr_reader :params
+  attr_reader :params, :user
 
   def scope
-    Submission.preload(:user, :domain)
+    Submission.preload(:user, :domain, :tags).then do |rel|
+      if user.present?
+        # hidden submissions aren't going to show up on the index so we don't have
+        # to select the action id here
+        rel.left_join_saved_info_for(user).where(hidden_submissions.arel.exists.not).
+          select("submissions.*, NULL AS hidden_action_id, saved_actions.id AS saved_action_id")
+      else
+        rel.select("submissions.*, NULL AS hidden_action_id, NULL AS saved_action_id")
+      end
+    end
+  end
+
+  def hidden_submissions
+    user.
+      submission_actions.hidden.
+      where("submission_actions.submission_short_id = submissions.short_id").
+      select(1)
   end
 
   def pagination_params
--- a/db/migrate/20200516221059_create_submission_actions.rb
+++ b/db/migrate/20200516221059_create_submission_actions.rb
@@ -1,0 +1,18 @@
+class CreateSubmissionActions < ActiveRecord::Migration[6.0]
+  def change
+    create_table :submission_actions do |t|
+      t.belongs_to :user, null: false, index: false, foreign_key: { on_delete: :cascade }
+      t.string :submission_short_id, null: false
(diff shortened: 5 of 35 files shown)


hernanat:
[As a user I can save / hide a submission](https://trello.com/c/iNwsvA8X/31-as-a-user-i-can-save-hide-a-submission)

app/controllers/submissions_controller.rb
```diff
@@ -24,6 +24,23 @@ def create
 
   private
 
+  def submission
+    Submission.friendly.preload(:user, :domain).then { |rel|
```
hernanat, line 28: yeah so this query (and similarly the one in `SubmissionSearch`)...and the way that we use the results is sort of odd.

Honestly this is one of the issues with ORMs -- they try to map relational concepts which are inherently dynamic onto static object models.

What I mean by this is that, in SQL, whenever you write a query, you are defining a relation. The way that ORMs work is that for a given model, they assume that the relation is always going to look something like:

```sql
SELECT * FROM model_table;
```

with maybe some filtering attached to that.

When you start wanting to use more complicated relations (like in our case how we want to return certain context-specific information for each submission as it relates to the current user), this starts to break down. `hidden_action_id` and `saved_action_id` are not attributes defined on the `Submission` model (and they shouldn't be, since their meaning is entirely user context-specific); but `ActiveRecord` will actually define them in in a hidden way since we're actually selecting them here (which you wouldn't know unless you kept this in mind), and so we can use them in this way.

We have to be careful while doing this, otherwise we'll get some  weird behaviors. I've tested this pretty thoroughly, and as you''ll see in the submission helper spec, I actually expect that for a regular submission object, certain methods will raise an error. That way we won't get things like unexpected UI bugs because these methods will fail loudly (and force us to make a decision).

app/models/concerns/short_id.rb
```diff
@@ -13,7 +13,7 @@ def set_short_id
     short_id = nil
 
     loop do
-      short_id = SecureRandom.urlsafe_base64(6).gsub(/-|_/, ("a".."z").to_a[rand(26)])
+      short_id = "#{self.class.short_id_prefix}#{SecureRandom.base36(8)}"
```
hernanat, line 16: I'm changing this to use base36 so that we don't have to include capital letters in short ids which can lead to some ambiguity. I also introduced an `s_` prefix in the case of submissions (and any class that wishes to use a short id must define a prefix). This will leave us with over 10^32 possible short ids for each prefix still, and I'm sure we'll all be long gone before we ever hit that.

app/models/submission_action.rb
```diff
@@ -0,0 +1,9 @@
+class SubmissionAction < ApplicationRecord
```
hernanat, line 1: lobsters breaks these out into two separate tables but since these are relatively similar concepts (i.e. a user does a thing to a submission) I figured that we can keep them in one table. Plus, if we want to add more actions, we don't have to worry about adding more tables.

app/search_objects/submission_search.rb
```diff
@@ -12,10 +15,26 @@ def results_paginator
...
+  attr_reader :params, :user
 
   def scope
-    Submission.preload(:user, :domain)
+    Submission.preload(:user, :domain, :tags).then do |rel|
+      if user.present?
+        # hidden submissions aren't going to show up on the index so we don't have
+        # to select the action id here
+        rel.left_join_saved_info_for(user).where(hidden_submissions.arel.exists.not).
+          select("submissions.*, NULL AS hidden_action_id, saved_actions.id AS saved_action_id")
```
hernanat, line 26: all of the `hidden_action_id`s are implicitly `nil` on the submission index where this search object is used (since we're not displaying hidden submissions there!)

db/migrate/20200516221059_create_submission_actions.rb
```diff
@@ -0,0 +1,18 @@
...
+    create_table :submission_actions do |t|
+      t.belongs_to :user, null: false, index: false, foreign_key: { on_delete: :cascade }
+      t.string :submission_short_id, null: false
+      t.integer :kind, null: false
+
+      t.timestamps
+    end
+
+    add_foreign_key :submission_actions, :submissions,
+      column: :submission_short_id, primary_key: :short_id
```
hernanat, line 12: once we get this into production and we have people actually start using the application, I'm going to revisit these indexes and see which ones are actually getting used by the query planner, and drop the ones that aren't. I suspect that the triplet unique index will be the most heavily used, but you can't really know until your production database has been running queries for a bit and has collected statistics and things for the query planner.

db/seeds.rb
```diff
@@ -5,3 +5,13 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+User.admin.
```
hernanat, line 9: the seeds file can be run by running `bundle exec rake db:seed`

spec/helpers/submissions_helper_spec.rb
```diff
@@ -52,4 +51,86 @@ module SubmissionsHelper
...
+        expect(helper.save_link_text_for(submission)).
+          to eq(t("submissions.submission_list_item.actions.default.saved"))
+      end
+    end
+
+    context "when `saved_action_id` is not defined on the submission" do
+      it "raises an error" do
+        submission = build(:submission, :text)
+
+        expect { helper.saved_link_text_for(submission) }.to raise_error(NoMethodError)
```
hernanat, line 104: again you'll notice that I'm expecting an error to be raised. The alternative approach would have been to check if `saved_action_id` (or `hidden_action_id`) was defined on the submission row, and if it wasn't then to just return the default or something, but then this could lead us to generating a ton of UI bugs (i.e. the action links having the wrong text). The idea is that if you are creating a page where you need to display action links, then you need to make sure that these fields are defined on the query relation.

spec/rails_helper.rb
```diff
@@ -47,11 +57,17 @@
   end
 
   config.before(:each, type: :system, js: true) do
-    driven_by(:selenium_chrome_headless)
+    driven_by(:chrome)
   end
-end
 
-Capybara.javascript_driver = :selenium_headless_chrome
+  config.around(:each, type: :system) do |example|
```
hernanat, line 63: I spent nearly 5 hours trying to figure out what the fuck was going on with capybara / chrome yesterday. It was failing to trigger a request when the action links were being clicked. Finally discovered that it wasn't really a capybara / chrome issue at all, but the fact that for whatever reason, the rails test environment sets

```ruby
ActionController::Base.allow_forgery_protection = false
```

by default, which was screwing with our requests since I was including the CSRF token 🤷

spec/system/submissions_index_spec.rb
```diff
@@ -77,4 +77,148 @@
       end
     end
   end
+
+  context "when a submission was hidden by a user" do
```
hernanat, line 81: for here and the view page, I really tried to test the hell out of this

hernanat:
action results on index page on click:

Screen Shot 2020-05-17 at 1 43 09 PM

action results if you go to a submission view page:

Screen Shot 2020-05-17 at 1 43 58 PM

ashpil:
Looks good! I do also read your comments for the backend, although I don't understand any of it.

I am sort of annoyed with how the `|` to the left of saved becomes green as well. Not sure how easy that would be to fix with the current implementation, though. Also I think if we're highlighting 'saved' when the user saved a post, we should also highlight 'hidden' when a user hides it.

hernanat:
> I am sort of annoyed with how the | to the left of saved becomes green as well.

I didn't even realize that; it's because the `|` gets added in an`:after` pseudo class. That should be easy enough to fix.

> Also I think if we're highlighting saved when the user saved a post, we should also highlight hidden when a user hides it.

Should we highlight it or make the text more transient?

ashpil:
I think even italicizing it or something like that would look fine.

hernanat:
Screen Shot 2020-05-17 at 5 45 20 PM
Screen Shot 2020-05-17 at 5 45 28 PM

hows this @ashpil ?

ashpil:
Looks good! Not 100% set on the italic, but better than nothing.

hernanat:
we can iterate :) hehe

hernanat:
thanks @ashpil !

PR #19, opened 2020-05-18
adds comment model and some supporting functionality

hernanat:
https://trello.com/c/Jj1W2US4

This commit introduces a `Comment` model for holding comments
  on submissions.

We also introduce the an extension for postgres called
  [LTREE](https://www.postgresql.org/docs/current/ltree.html).

This extension is used to support the tree structure that arises
  from comments.

Since we're planning to use postgres in production, and my focus
  is not to try to dive into the weeds in MySQL and other DBMs,
  I've deferred adding support for non-postgres DBMs in this commit.
  (Though if anyone feels like it, I've purposefully structured this to
  be extendable).

### DIFF
--- a/app/models/comments/postgresql_tree_adapter.rb
+++ b/app/models/comments/postgresql_tree_adapter.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Comments
+  module PostgresqlTreeAdapter
+    extend ActiveSupport::Concern
+
+    included do
+      before_create :set_ancestry_path
+
+      has_many :children, class_name: "Comment", foreign_key: :parent_id
+    end
+
+    def root_node?
+      parent_id.nil?
+    end
+
+    def childless?
+      children.none?
+    end
+
+    # the `@>` operator is asking "is the left an ancestor of the right?"
+    def descendants
+      self.class.where("? @> ancestry_path", ancestor_path)
+    end
+
+    def siblings
+      # if the ancestry path is present, we want to find all nodes
+      # with the same ancestry path. if it's not, then we have a
+      # root node which cannot have siblings.
+      # we exclude `self` from the resulting relation.
+      if ancestry_path.present?
+        self.class.where("? = ancestry_path", ancestry_path).where.not(id: id)
+      else
+        self.class.none
+      end
+    end
+
+    private
+
+    def set_ancestry_path
+      if parent.present?
+        prefix = (parent.ancestry_path.present? ? "#{parent.ancestry_path}." : "")
+        self.ancestry_path = "#{prefix}#{parent.short_id}"
+      end
+    end
+
+    def ancestor_path
+      prefix = ancestry_path.present? ? "#{ancestry_path}." : ""
+      "#{prefix}#{short_id}"
+    end
+  end
+end
--- a/db/migrate/20200517210826_create_comments.rb
+++ b/db/migrate/20200517210826_create_comments.rb
@@ -1,0 +1,28 @@
+class CreateComments < ActiveRecord::Migration[6.0]
+  def change
+    adapter = ActiveRecord::Base.connection.adapter_name.downcase
+    enable_extension "ltree" if adapter == "postgresql"
+    create_table :comments do |t|
+      t.string :short_id, null: false
+      t.belongs_to :user, null: false, foreign_key: { on_delete: :cascade }
+      t.belongs_to :submission, null: false, foreign_key: { on_delete: :cascade }
+      # we're keeping the reference to parent to enforce some degree of foreign
+      # key constraints. We don't want to end up with orphaned trees.
+      t.belongs_to :parent, null: true,
+        foreign_key: { to_table: :comments, on_delete: :cascade }
+      t.text :body, null: false
+      if adapter == "postgresql"
+        # use LTREE extension for efficiently managing comment ancestry paths
+        t.ltree :ancestry_path, null: true
+      else
+        # TODO: figure out if useful extensions exist for other dbms
+        t.text :ancestry_path, null: true
+      end
+
+      t.timestamps
+    end
+
+    add_index :comments, :short_id, unique: true
+    add_index :comments, :ancestry_path, using: :gist if adapter == "postgresql"
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,10 +10,27 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_05_16_221059) do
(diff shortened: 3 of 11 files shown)


app/models/comments/postgresql_tree_adapter.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module Comments
+  module PostgresqlTreeAdapter
```
hernanat, line 4: the idea is that for other DBMs, they're similarly implement their own adapter. I have no plans on implementing any myself but at least here people can if they want to.

app/models/comments/postgresql_tree_adapter.rb
```diff
@@ -0,0 +1,52 @@
...
+      if ancestry_path.present?
+        self.class.where("? = ancestry_path", ancestry_path).where.not(id: id)
+      else
+        self.class.none
+      end
+    end
+
+    private
+
+    def set_ancestry_path
```
hernanat, line 40: it's worth mentioning that the canonical way of thinking of the ancestry path is by means of a reflexive + transitive (but not symmetric) relation -- i.e.

If we have nodes `A`, `B`, `C`, and if `B` is a child of `A`, and `C` is a child of `B`, then:

- `A` has ancestry path `A`
- `B` has ancestry path `A.B`
- `C` has ancestry path `A.B.C`

we get:

- `A @> A = true` (`A` is an ancestor of itself)
- `A @> A.B = true` (`A` is an ancestor of `B`)
- `A.B @> A.B.C = true` (`B` is an ancestor of `C`)
- `A @> A.B.C = true` (`A` is an ancestor of `C`)

The way I am setting the ancestry path here _does not_ follow this convention. The reason being that, in this context, when we say "get me the descendants of `A`", we don't want to also include `A` in that (because we're starting from `A` already).

I could have accomplished this by just tacking on an additional `where` condition to ignore `A` itself in the query, but that felt awkward. Plus, if we wanted to do it that way, we would also have to then care about the order in which the `before_create` callbacks get executed on a comment record -- recall that `short_id` gets assigned `before_create`, and so if for whatever reason the order of the callbacks switched and this was triggered before the `short_id` setter, then this would break.

app/models/comments/postgresql_tree_adapter.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module Comments
+  module PostgresqlTreeAdapter
+    extend ActiveSupport::Concern
+
+    included do
+      before_create :set_ancestry_path
+
+      has_many :children, class_name: "Comment", foreign_key: :parent_id
```
hernanat, line 10: I have the parent foreign key on the comments table anyways (we keep a foreign key constraint to prevent orphaned comments), so we don't need to use the LTREE stuff to get the direct children.

app/models/comments/postgresql_tree_adapter.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def root_node?
```
hernanat, line 13: these are all the methods I could think to implement at this point. I'm sure others will come up.

db/migrate/20200517210826_create_comments.rb
```diff
@@ -0,0 +1,28 @@
...
+    create_table :comments do |t|
+      t.string :short_id, null: false
+      t.belongs_to :user, null: false, foreign_key: { on_delete: :cascade }
+      t.belongs_to :submission, null: false, foreign_key: { on_delete: :cascade }
+      # we're keeping the reference to parent to enforce some degree of foreign
+      # key constraints. We don't want to end up with orphaned trees.
+      t.belongs_to :parent, null: true, index: false,
+        foreign_key: { to_table: :comments, on_delete: :cascade }
+      t.text :body, null: false
+      if adapter == "postgresql"
```
hernanat, line 14: I know that this is ugly but that's the nature of what we're dealing with since there is desire to allow folks to swap out dbms.

db/migrate/20200517210826_create_comments.rb
```diff
@@ -0,0 +1,28 @@
...
+      else
+        # TODO: figure out if useful extensions exist for other dbms
+        t.text :ancestry_path, null: true
+      end
+
+      t.timestamps
+    end
+
+    add_index :comments, :short_id, unique: true
+    add_index :comments, :ancestry_path, using: :gist if adapter == "postgresql"
```
hernanat, line 26: the `ltree` datatype supports a gist index which `@>` operation (and others) can benefit from.

db/schema.rb
```diff
@@ -10,11 +10,28 @@
...
+    t.string "short_id", null: false
+    t.bigint "user_id", null: false
+    t.bigint "submission_id", null: false
+    t.bigint "parent_id"
+    t.text "body", null: false
+    t.ltree "ancestry_path"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["ancestry_path"], name: "index_comments_on_ancestry_path", using: :gist
+    t.index ["parent_id"], name: "index_comments_on_parent_id"
```
hernanat, line 29: I'm thinking about whether or not it would make more sense to ditch the `parent_id` index (but keep the foreign key -- or ditch the foreign key? I don't know. I worry about things getting orphaned but how likely is that really? 🤷 ) and just derive this information using the LTREE information. Eh. Can always revisit it later.

app/models/comments/postgresql_tree_adapter.rb
```diff
@@ -0,0 +1,52 @@
...
+      parent_id.nil?
+    end
+
+    def childless?
+      children.none?
+    end
+
+    # the `@>` operator is asking "is the left an ancestor of the right?"
+    def descendants
```
hernanat, line 22: one of the things that bothers me about what we've done here is that there isn't an obvious way to preload these relations in order to prevent `N+1` queries. I spent some time in the `ActiveRecord` internals earlier trying to figure out a way to do this, but I feel like it may be something that I will have to come back to later. Rails _really_ doesn't want you to do stuff like this so I imagine there isn't an obvious public API exposed anywhere. Many of the search results I came across from other folks asking questions were mostly related to preloading scopes, which is not what we want to do here.

In any case I don't think it will be too bad of a performance hit since we'll only be query comments per-submission anyways. Still, I'm bothered.

hernanat:
@ashpil this is just backend stuff for supporting comments + the tree structure that arises from them. Not sure if you want to / are interested in going through it ~ otherwise I'll just merge it in and start working on the user facing part of comments tomorrow :) (by the way if you want to work on any part of that let me know -- i.e. submitting comments, displaying comments, etc.)

PR #24, opened 2020-05-20
query improvements

hernanat:
this commit introduces two database views for
  helping improve query performance to speed up
  rendering:

- `FlattenedSubmission` which contains all of the necessary
  information for displaying a submission

- `FlattenedComment` similar to above

This prevents us from having to load supporting objects alongside
  the submissions and comments because we just query for the fields
  that we are interested in.


Gemfile
```diff
@@ -1,44 +1,45 @@
-source 'https://rubygems.org'
+source "https://rubygems.org"
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-ruby '2.7.1'
+ruby "2.7.1"
 
+gem "bitters"
```
hernanat, line 6: I should have did this as a separate PR but 🤷 I'm just adding double quotes to keep things consistent

Gemfile
```diff
@@ -1,44 +1,45 @@
...
 gem "devise-i18n"
 gem "faraday"
+gem "friendly_id", "~> 5.2.4"
 gem "haml-rails", "~> 2.0"
+gem "jbuilder", "~> 2.7"
+gem "pg", ">= 0.18", "< 2.0"
+gem "puma", "~> 4.1"
+gem "rails", "~> 6.0.3"
+gem "sass-rails", ">= 6"
+gem "scenic"
```
hernanat, line 20: this is the more important change -- scenic is a gem that allows us to easily use database views / materialized views.

app/search_objects/flattened_submission_search.rb
```diff
@@ -0,0 +1,64 @@
...
+class FlattenedSubmissionSearch
+  DEFAULT_PAGE = 0
+  DEFAULT_PER_PAGE = 25
+
+  def initialize(params, user)
+    @params = params
+    @user = user
+  end
+
+  def by_short_id(short_id)
```
hernanat, line 12: this either belongs somewhere else, or we need to rename this class. We want some of the same information (namely the action information) for a single submission as we do with a list of them from the context of a given user. Or maybe this additional wrapper information needs to get tacked on in a different place and doesn't belong as part of the search. I'll have to think about that.

app/views/comments/_comment_tree.html.erb
```diff
@@ -0,0 +1,42 @@
+<% subtree = comments_by_parent[nil] %>
```
hernanat, line 1: this is fucking horrendous and I'm going to have to come back to it when my brain is working again.

Basically, we want to iteratively build the comment tree in the UI, because in instances where we have a sufficiently large number of comments, building the tree recursively (i.e. the easy way) will blow our stack up and we don't want that happening.

I'm also trying to figure out ways to speed up the rendering. In my testing I noticed that once we hit around ~60 comments, returning a page starts taking around ~500ms, with over 95% of that time being taken up by rendering the page itself (it makes me happy that very little time is taken in the database :))

for example, on a page with 62 comments:

```
Completed 200 OK in 505ms (Views: 484.2ms | ActiveRecord: 8.0ms | Allocations: 404807)
```

I did some playing around on Reddit it and this appears to be on par with their page load times. But it still makes me sad.

I think reducing the amount of actual computations being done in the partial here would help, but also it might be worth exploring something like infinite scroll in the future, where we only load a subset of comments at a time. This would (maybe?) make for a better user experience, and would also free up our server from having to render huge amounts of HTML at once (which becomes a performance problem given a significant number of requests).

Another option still is to explore what we can do with caching, though I'm not sure what our options are there given that the nature of these comment sections are that they are dynamic entities.

db/views/flattened_comments_v01.sql
```diff
@@ -0,0 +1,6 @@
+SELECT
```
hernanat, line 1: this one is puny at the moment, but the usefulness of it will become clearer when we introduce voting

db/views/flattened_submissions_v01.sql
```diff
@@ -0,0 +1,32 @@
...
+      WHERE comments.submission_id = submissions.id
+      GROUP BY submission_id
+    ), 0) AS comment_count
+FROM
+  submissions
+INNER JOIN users ON submissions.user_id = users.id
+INNER JOIN domains ON submissions.domain_id = domains.id
+WHERE submissions.domain_id IS NOT NULL
+
+UNION ALL
```
hernanat, line 17: I'm using `UNION ALL` because the nature of how we allow submissions to be created is that there can't be overlap between submissions with URLs and submissions with text bodies, so we don't need the additional de-duplication step that comes with a regular `UNION`, which is a performance win.

app/views/comments/_comment_tree.html.erb
hernanat, line 1: also:

```erb
<% subtree = comments_by_parent[nil] %>
```

here we're using the fact that root comments will have `nil` parents

ashpil: Doesn't Reddit load only parts of the comments? Like there's a "load more comments" button when the tree gets either very deep or broad.

hernanat: yup! they do infinite scroll, though they're typically dealing with > 100 comments on average. I'm not sure how frequently we'd expect that on our threads, so I'd rather punt on it and make some minor tweaks in the meantime if we can (it's a pain in the ass to implement so it's the sort of thing where I'd rather wait until the very last second where we need it 😇 )

hernanat:
thanks @ashpil !

PR #26, opened 2020-05-20
Added styling for alerts

ashpil:
This makes the alerts a hide-able, unobtrusive popup:

demo

(There's a shadow to make it more clear there's a popup but it doesn't show very will in the gif)

https://trello.com/c/L7Ga9Sr5

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -11,8 +11,11 @@
       = yield :page_styles
   %body
     = render "layouts/navbar"
-    - flash.each do |key, value|
-      .alert{ class: "alert-#{key}" }
-        = value
+    - flash.each do |flash_type, flash_message|
+      %label.flash_container
+        %input{ type: "checkbox" }
+        .flash{ class: "flash-#{flash_type}" }
+          = flash_message
+          %span.flash_close x
     .application-container
       = yield
--- a/app/assets/stylesheets/base/_base.scss
+++ b/app/assets/stylesheets/base/_base.scss
@@ -13,3 +13,4 @@
 @import "tables";
 @import "typography";
 @import "markdown";
+@import "flash";
--- a/app/assets/stylesheets/base/_flash.scss
+++ b/app/assets/stylesheets/base/_flash.scss
@@ -1,0 +1,22 @@
+.flash_container {
+    cursor: pointer;
+    .flash {
+        position: fixed;
+        padding: $spacing-small;
+        background-color: $white;
+        box-shadow: $box-shadow;
+        top: 2rem;
+        left: 50%;
+        transform: translateX(-50%);
+        font-weight: normal;
+        .flash_close {
+            font-family: $font-stack-mono;
+            color: lighten($scorpion, 20%);
+            margin-left: $spacing-small;
+        }
+    }
+    input[type="checkbox"],
+    :checked + .flash {
+        display: none;
+    }
+}
--- a/spec/support/page_objects/page_base.rb
+++ b/spec/support/page_objects/page_base.rb
@@ -14,6 +14,6 @@ class PageBase
   end
 
   def has_notice?(notice)
-    has_css?(".alert-notice", text: notice)
+    has_css?(".flash-notice", text: notice)
   end
 end


ashpil:
Now that I've submitted this PR, I just realized I'm not sure how accessible this method is/how to make it better.

app/views/layouts/application.html.haml
```diff
@@ -10,7 +10,11 @@
...
     = render "layouts/navbar"
     - flash.each do |key, value|
-      .alert{ class: "alert-#{key}" }
-        = value
+      %label{ class:"alert_container"}
+        %input{:type=>"checkbox", id: "alert_checkbox"}
+        .alert{ class: "alert-#{key}" }
+          = value
+          %span{ class:"alert_close"}
+            = "x"
```
ashpil, line 18: Uh-oh, CircleCI doesn't like something here. Once again I'm unsure how to decrypt i's message. I wasn't sure how exactly proper haml formatting worked, so I kind of followed the rest of the file but apparently not in the way I was supposed to.

ashpil, line 18: Uh-oh, CircleCI doesn't like something here. Once again I'm unsure how to decrypt i's message. I wasn't sure how exactly proper haml formatting worked, so I kind of followed the rest of the file but apparently not in the way I was supposed to.

ashpil:
Damn CircleCI is opinionated. I love it.

I think I fixed them all, [this](https://github.com/sds/haml-lint/blob/master/lib/haml_lint/linter/README.md) page is a godsend.

ashpil:
[Style the alert](https://trello.com/c/L7Ga9Sr5/33-style-the-alert)

hernanat:
you can also run:
`bundle exec rubocop -a`
and
`bundle exec haml-lint app/views/**/*.haml`

locally so you don't have to wait for CI to run. The rubocop one will actually try to correct formatting issues automatically (the `-a` flag) which is helpful

ashpil:
> locally so you don't have to wait for CI to run

Sweet, I figured there was a way to do that. I'll do those before committing next time.

app/views/layouts/application.html.haml
```diff
@@ -10,7 +10,10 @@
   %body
     = render "layouts/navbar"
     - flash.each do |key, value|
-      .alert{ class: "alert-#{key}" }
-        = value
+      %label.alert_container
+        %input{ type: "checkbox" }
+        .alert{ class: "alert-#{key}" }
```
hernanat, line 15: just some terminology here, and I realize that I'm actually the one who committed the initial version of this so shame on me (must have been another sleepless night):

there are two usual flavors of flash messages in rails (though we can define however many we want), `alert` and `notice`.

`alert` typically means something went wrong, whereas `notice` is usually just used to notify the user that whatever they tried to do went through.

what do you think about renaming the `.alert` class to `.flash` instead, and then use `"flash-#{key}"` for selecting the extra class? (funny enough right now we could end up with `alert-alert` haha)?

ashpil: Why `.flash` rather than `.notice`?

hernanat: `.flash` is the general term, so it would be something like this:

```haml
%label.flash_container
  %input{ type: "checkbox" }
  .flash{ class: "flash-#{key}" }
```

in the line above your changes, i.e. `flash.each do |key, value|`, `flash` is actual a hash defined in Rails for storing flash messages. The key is the flash "type" and the value is the message. So really it would probably read better if I had written it like this:

```haml
- flash.each do |flash_type, flash_message|
  ...etc
```

ashpil: If you wanna commit changes with that now, feel free, but it seems somewhat more technical than the rest of the stuff I’ve been doing, so if you don’t want to do it, I’ll do it in the morning when I have more time.

hernanat: yeah I can do that -- I'm probably going to sign off soon as well. I'm at the end of my patience with this font issue. Idk if you've had a chance to look at the app in heroku yet but there are square roots flying off the right side of the page into infinity -_____-

ashpil:
Wow, that commit is much simpler than what I thought you were describing, haha. 

Also, CircleCI fears no mortal.

hernanat:
> Also, CircleCI fears no mortal.

yeah ugh I'm an idiot. I accidentally moved the file up a directory when I renamed it --________--

hernanat:
> Now that I've submitted this PR, I just realized I'm not sure how accessible this method is/how to make it better.

hm. we probably need to just figure out a way to give it focus when it's present / make them tabbable. Not sure the best way to do that. Can look into it tomorrow.

ashpil:
> 
> 
> > Now that I've submitted this PR, I just realized I'm not sure how accessible this method is/how to make it better.
> 
> hm. we probably need to just figure out a way to give it focus when it's present / make them tabbable. Not sure the best way to do that. Can look into it tomorrow.

Yeah, the thing is that we've already given focus to the comment just posted, at least for when you get a notice that your comment was submitted. So there's multiple relevant things to focus there. Ideally we want the focus to go from the notice directly to the relevant comment. Not sure how to do that.

ashpil:
I'm unsure how to make the flashes immediately in focus, as I don't understand how the comment is focused. I think it's fine if we just push it for now and make it better later, or what do you think?

hernanat:
Yeah let’s do that for now. No need to get in the weeds on this one

ashpil:
Sorry, I had to merge and rebase so I need another approval now.

PR #28, opened 2020-05-20
fixes production LaTeX issue

hernanat:
I _despise_ webpacker. Basically, it was ignoring the imported
  katex styles in production.

Well...actually it was more so that in development it magically
  just worked even though we weren't using the stylesheet pack
  tag to import that styles related to the pack.

I'm too annoyed to try to figure out the reasoning behind this
  behavior at the moment.

### DIFF
--- a/app/javascript/packs/latex.js
+++ b/app/javascript/packs/latex.js
@@ -1,4 +1,4 @@
-import '../latex_styles';
+import '../latex_styles.sass';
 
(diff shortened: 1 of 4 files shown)


hernanat:
my blood pressure is so high right now

I've spent the entire day trying to figure this out.

hernanat:
at least now if anything like this happens in the future I know where to start T___T

PR #30, opened 2020-05-21
introduces `Vote` model for up/down votes

hernanat:
https://trello.com/c/w9pQI5lQ
https://trello.com/c/FiwnyH46
https://trello.com/c/DD0nQfG3
https://trello.com/c/Pb08HQGH

This commit just involves the model for storing information
  about voting on submissions / comments.

Subsequent commits will introduce the ability to vote, starting
  with upvoting.

### DIFF
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+class Vote < ApplicationRecord
+  belongs_to :votable, polymorphic: true
+  belongs_to :user
+  validate :supported_votable_type
+  validates :downvote_reason,
+    absence: { message: I18n.t("votes.errors.downvote_reason_for_upvote") },
+    if: ->(v) { v.upvote? }
+  validates :downvote_reason,
+    presence: { message: I18n.t("votes.errors.missing_downvote_reason") },
+    if: ->(v) { v.downvote? }
+  validate :valid_downvote_reason,
+    if: ->(v) { v.downvote? && v.downvote_reason.present? }
+
+  enum kind: {
+    upvote: 0,
+    downvote: 1
+  }
+
+  enum downvote_reason: {
+    off_topic: 0,
+    incorrect: 5,
+    spam: 10,
+    mean: 15,
+    unhelpful: 20,
+    troll: 25,
+    broken_link: 30,
+    repost: 35
+  }
+
+  SUBMISSION_DOWNVOTE_REASONS = %i[off_topic incorrect spam troll broken_link repost].freeze
+  COMMENT_DOWNVOTE_REASONS = %i[off_topic incorrect spam mean unhelpful troll].freeze
+  SUPPORTED_VOTABLE_TYPES = [Submission, Comment].freeze
+
+  private
+
+  def supported_votable_type
+    unless SUPPORTED_VOTABLE_TYPES.include?(votable_type.constantize)
+      errors.add(
+        :votable_type,
+        I18n.t("votes.errors.unsupported_votable_type", type: votable_type.pluralize.downcase)
+      )
+    end
+  end
+
+  def valid_downvote_reason
+    case votable
+    when Submission
+      validate_submission_downvote_reason
+    when Comment
+      validate_comment_downvote_reason
+    end
+  end
+
+  def validate_submission_downvote_reason
+    unless SUBMISSION_DOWNVOTE_REASONS.include?(downvote_reason.to_sym)
+      errors.add(:downvote_reason, I18n.t("votes.errors.invalid_submission_downvote_reason"))
+    end
+  end
+
+  def validate_comment_downvote_reason
+    unless COMMENT_DOWNVOTE_REASONS.include?(downvote_reason.to_sym)
+      errors.add(:downvote_reason, I18n.t("votes.errors.invalid_comment_downvote_reason"))
+    end
+  end
+end
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -5,6 +5,7 @@ class Comment < ApplicationRecord
   belongs_to :user
   belongs_to :submission
   belongs_to :parent, optional: true, class_name: "Comment"
+  has_many :votes, as: :votable
 
   validates :body, presence: true
 
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -7,6 +7,7 @@ class Submission < ApplicationRecord
   has_many :tags, through: :submission_tags
   has_many :comments
   has_many :root_comments, -> { where(parent_id: nil) }, class_name: "Comment"
+  has_many :votes, as: :votable
 
   def self.short_id_prefix
     :s_
--- a/config/locales/votes.en.yml
+++ b/config/locales/votes.en.yml
@@ -1,0 +1,17 @@
+en:
+  votes:
+    downvote_reason:
+      off_topic: "Off-topic"
+      incorrect: "Incorrect / Inaccurate"
+      spam: "Spam"
+      mean: "Mean-spirited / unfriendly"
+      unhelpful: "Unhelpful"
+      troll: "Trolling"
+      broken_link: "Broken link"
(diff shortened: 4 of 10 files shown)


hernanat:
[As a user I can upvote a submission](https://trello.com/c/w9pQI5lQ/4-as-a-user-i-can-upvote-a-submission)

hernanat:
[As a user I can upvote a comment](https://trello.com/c/FiwnyH46/5-as-a-user-i-can-upvote-a-comment)

hernanat:
[As a user I can downvote a submission](https://trello.com/c/DD0nQfG3/6-as-a-user-i-can-downvote-a-submission)

hernanat:
[As a user I can downvote a comment](https://trello.com/c/Pb08HQGH/7-as-a-user-i-can-downvote-a-comment)

app/models/vote.rb
```diff
@@ -0,0 +1,67 @@
...
+    incorrect: 5,
+    spam: 10,
+    mean: 15,
+    unhelpful: 20,
+    troll: 25,
+    broken_link: 30,
+    repost: 35
+  }
+
+  SUBMISSION_DOWNVOTE_REASONS = %i[off_topic incorrect spam troll broken_link repost].freeze
```
hernanat, line 32: I really like the idea of requiring a reason for downvoting (which is what lobsters does). In the future maybe we could even expand on this to give the option to give a brief summary alongside just selecting a reason so that authors can receive more detailed feedback?

ashpil: I like some sort of system like this as well, though I think the set of reasons for downvoting may be somewhat limiting. 

I think there might be a different better system that accomplishes the same thing, though. 

What do you think about only allowing users to downvote a submission/comment if they’ve replied to it? That way you can only downvote something if you’ve already written down for everyone to see what’s wrong/misinformed in a given post. 

And then also we can like reverse someone’s downvote if their explanatory comment goes to -3 or something.

ashpil: I like the reasoning for downvoted system, but I feel like on Reddit it even lobsters you stumble onto a post with -5 and you don’t even know why it’s at -5. This would insure everyone knows what people think is wrong about a given post, and allow people to judge for themselves whether they think the downvotes are justified.

hernanat: > though I think the set of reasons for downvoting may be somewhat limiting

are you suggesting we nix the enumerated set of reasons altogether, and require them to write a reason why, or do you have more that you would include? Or maybe we add an "other" section?

> What do you think about only allowing users to downvote a submission/comment if they’ve replied to it? That way you can only downvote something if you’ve already written down for everyone to see what’s wrong/misinformed in a given post.
> And then also we can like reverse someone’s downvote if their explanatory comment goes to -3 or something.

I like this a lot. My only critique of that would be in the case where something is very clearly a spam / troll post, we shouldn't require everyone to have to post something as well as select a reason for downvoting if that makes sense. We can definitely create different rulesets for different situations, though.

Let's make a card for this and when I get to actually implementing the downvoting mechanism I'll take this approach (I'm doing upvotes first because they're easy)

hernanat: > I like the reasoning for downvoted system, but I feel like on Reddit it even lobsters you stumble onto a post with -5 and you don’t even know why it’s at -5. This would insure everyone knows what people think is wrong about a given post, and allow people to judge for themselves whether they think the downvotes are justified.

Maybe we can add some sort of tag that gets automatically applied that is a stand in for the most used downvote reason/

For example, if someone posts something that is offtopic, and the downvotes exceed some threshhold (or the score goes negative), we calculate what the most used downvote reason is and apply it as a tag to the votable entity.

ashpil: > are you suggesting we nix the enumerated set of reasons altogether

Yes.

> My only critique of that would be in the case where something is very clearly a spam / troll post, we shouldn't require everyone to have to post something as well as select a reason for downvoting if that makes sense.

I think we may actually be able to benefit from this. We can clearly separate "downvotable offenses" and "reportable offenses", which I think makes a downvote more meaningful. If something is clearly spam, troll, or even maybe off-topic, that's a reportable offense. You can click a report button and select which one of those it is. That goes to the moderator, because something like that doesn't really benefit from any sort of discussion.

If something is incorrect or unhelpful, or even something more nuanced, that's a downvotable post. Then, if you'd like to downvote something, you'd have to reply to it, and in a way start a discussion and let others know what is incorrect, which can't really be done properly in a unnuanced one word reason, even if the public gets to see it. 

This way, a downvote is purely about discussion, rather than just about moderating things you don't like out of the spotlight (eg. spam). It separates relevancy and quality.

hernanat: you raise good points -- I'll have a PR up in a few minutes

hernanat: do you think that we should also make some sort of distinction in the UI for users that downvote something? that seems like it could potentially be singling people out though. Also how would we distinguish downvote reasons from other comments they might make in the thread?

hm.

Maybe when they fill out the downvote reason form itself it triggers a comment, and that comment could be flagged as a downvote reason.

In the UI, maybe we only make a distinction between downvoters to the mods / admins...this could help prevent people from diving off into flamewars.

Do you have any ideas?

ashpil: I didn't think of distinguishing it to such an extent.

I figured if a new user wanted to downvote something it would go like so:
1. They press the downvote button, and
2. Get a flash that says something like, "Sorry, to downvote something you must first reply to it explaining your reasoning."
3. If the user still thinks the downvote is worth it, they reply to it, which acts completely like a normal comment. I figure it would typically be obvious that the user disagrees with the parent comment/submission and is planning to downvote it, as the user would say something like "This is incorrect because...." or "This is unhelpful because...". I don't think we have to implement any sort of special feature here, just have `COMMENT` as a prerequisite for `DOWNVOTE`. 
4. Now since the user has replied to it, they are able to downvote it. 

> Also how would we distinguish downvote reasons from other comments they might make in the thread?

So what I'm saying is that a `downvote reason` would actually be a normal comment, because wouldn't it serve the same purpose as one? 

Or do you think we should distinguish downvote reasons a bit more?

hernanat: I think for moderation purposes we probably want to (at least from our perspective) be able to distinguish between a downvote explanation comment and a regular comment. Id the thread ends up having 100 comments and a particular downvoter is involved in the conversation to varying degrees, it could be difficult to figure out what the initial reasoning was. So I guess what Im saying is that the downvote reason box could actually just be a comment box and when it gets submitted the comment has a special “downvote reason” flag or something.

hernanat: So instead of a comment being a prerequisite, it actually is submitted alongside the downvote

hernanat: There’s the obvious case where the explanation is submitted before the comment, but we can probably:

- rely on the fact that after a while users are conditioned to know that their reason is submitted alongside the comment

- give them the option to select an existing comment in the thread

ashpil: Hm. I guess that sort of system feels sort of unappealing to me, as it feels like unnecessary complexity. 

I suppose that's because I think if we do something like that, we're complicating it for the end user. In my opinion, just saying "If you want to downvote something, please first post a comment about what you dislike about it" is a lot simpler and more approachable than "Downvotes are submitted alongside downvote reasons, which are treated as comments."

> I think for moderation purposes we probably want to (at least from our perspective) be able to distinguish between a downvote explanation comment and a regular comment.

I agree with this. But, we can certainly achieve the same thing without explicitly making it like that on the front-end, for the user. We can see when someone replied to a submission, and when someone downvoted a submission, and thus, the reply someone posted immediately before downvoting.

hernanat: that's fair -- and I do think that the argument for simplicity is compelling. We can try out your suggestion for now and if it becomes an issue in the future we can always revisit our approach.

ashpil: Sounds good! 😄 I'm always down for revisting.

PR #31, opened 2020-05-21
simplifies downvoting vision

hernanat:
see discussion on https://github.com/hernanat/simple_group/pull/30

we've opted to keep downvoting simpler for now and not require
  users to select a predefined reason, which may be seen as limiting.

instead, we'll require that the user has already commented in response
  to what they're downvoting (behavior to come in a subsequent PR).

### DIFF
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -4,33 +4,12 @@ class Vote < ApplicationRecord
   belongs_to :votable, polymorphic: true
   belongs_to :user
   validate :supported_votable_type
-  validates :downvote_reason,
-    absence: { message: I18n.t("votes.errors.downvote_reason_for_upvote") },
-    if: ->(v) { v.upvote? }
-  validates :downvote_reason,
-    presence: { message: I18n.t("votes.errors.missing_downvote_reason") },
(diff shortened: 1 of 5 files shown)


hernanat:
Screen Shot 2020-05-21 at 1 11 07 PM
😎

hernanat:
I'll have to remember to make sure that if a user deletes their comments from a thread their downvote is also removed so that people don't try to game the system. This should be easy enough.

ashpil:
> Screen Shot 2020-05-21 at 1 11 07 PM
> 😎

Oof. Sorry bout all that extra code you wrote that's now all for naught.

hernanat:
all set with this @ashpil ?

ashpil:
> I'll have to remember to make sure that if a user deletes their comments from a thread their downvote is also removed so that people don't try to game the system. This should be easy enough.

Yeah, what I like about the comment system is I feel like there's a couple interesting ways we can extend it. (For example, as I mentioned before, we can remove a downvote from a parent comment if a subcomment from a user that downvoted the parent comment gets downvoted enough).

hernanat:
woops sorry I forgot to cleanup the factories

PR #32, opened 2020-05-21
fixes unique index on votes

hernanat:
I'm a dummy. Previously the unique index included the
  vote `kind` (i.e. `upvote` / `downvote`).

This is not desirable because then a user could conceivably
  have both an upvote and a downvote for a given votable
  entity. We only want to allow one or the other.


hernanat:
thanks @ashpil !

PR #33, opened 2020-05-22
adds in voting APIs for comments and submissions

hernanat:
https://trello.com/c/w9pQI5lQ
https://trello.com/c/FiwnyH46
https://trello.com/c/DD0nQfG3
https://trello.com/c/Pb08HQGH

this commit adds in the relevant upvote and downvote APIs
  for submissions and comments. The views will be hooked into
  them in a subsequent commit.

I opted to go with separate API endpoints rather than a single
  voting endpoint to avoid having to introduce additional logic
  to deal with an incoming vote request.

Plus, if for example, a request is made to the comment upvotes
  endpoint, we know for a fact that:

  - it is a request relating to a comment
  - it is some form of request relating to an upvote

Contrast this with if we had a single vote API (or even a vote
  API for each votable type), where we would have to include
  additional information in the request to try to tell the controller
  how to handle the request.

Under the scheme I've implemented, the only information that we need
  is the short id for the resource (and the currently logged in user).

### DIFF
--- a/app/controllers/api/v1/comments/downvotes_controller.rb
+++ b/app/controllers/api/v1/comments/downvotes_controller.rb
@@ -1,0 +1,60 @@
+module Api
+  module V1
+    module Comments
+      class DownvotesController < BaseController
+        def update
+          find_or_initialize_vote.then do |vote|
+            if comment.children.where(user_id: current_user.id).exists?
+              handle_vote(vote)
+            else
+              render json: build_response(:rejected).
+                merge(info: t(".comment_before_downvote")),
+                status: :forbidden
+            end
+          end
+        end
+
+        private
+
+        def handle_vote(vote)
+          if vote.persisted?
+            handle_existing_vote(vote)
+          else
+            handle_downvote(vote)
+          end
+        end
+
+        def handle_existing_vote(vote)
+          if vote.downvote?
+            vote.destroy!
+            render json: build_response(:removed), status: :ok
+          else
+            handle_downvote(vote)
+          end
+        end
+
+        def handle_downvote(vote)
+          vote.downvote!
+          render json: build_response(:downvoted), status: :ok
+        end
+
+        def build_response(action)
+          {
+            short_id: params[:comment_short_id],
+            action: action
+          }
+        end
+
+        def comment
+          @_comment ||= Comment.friendly.find(params[:comment_short_id])
+        end
+
+        def find_or_initialize_vote
+          current_user.votes.comment.find_or_initialize_by(votable_id: comment.id)
+        rescue ActiveRecord::RecordNotFound
+          head :not_found
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/comments/upvotes_controller.rb
+++ b/app/controllers/api/v1/comments/upvotes_controller.rb
@@ -1,0 +1,47 @@
+module Api
+  module V1
+    module Comments
+      class UpvotesController < BaseController
+        def update
+          find_or_initialize_vote.then do |vote|
+            if vote.persisted?
+              handle_existing_vote(vote)
+            else
+              handle_upvote(vote)
+            end
+          end
+        end
+
+        private
+
+        def handle_existing_vote(vote)
+          if vote.upvote?
+            vote.destroy!
+            render json: build_response(:removed), status: :ok
+          else
+            handle_upvote(vote)
+          end
+        end
+
+        def handle_upvote(vote)
+          vote.upvote!
+          render json: build_response(:upvoted), status: :ok
+        end
+
+        def build_response(action)
+          {
+            short_id: params[:comment_short_id],
+            action: action
+          }
+        end
+
+        def find_or_initialize_vote
+          comment = Comment.friendly.find(params[:comment_short_id])
+          current_user.votes.comment.find_or_initialize_by(votable_id: comment.id)
+        rescue ActiveRecord::RecordNotFound
+          head :not_found
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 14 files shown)


app/controllers/api/v1/comments/downvotes_controller.rb
```diff
@@ -0,0 +1,60 @@
+module Api
+  module V1
+    module Comments
+      class DownvotesController < BaseController
+        def update
+          find_or_initialize_vote.then do |vote|
```
hernanat, line 6: you'll notice that some of the logic across these controllers could have probably been DRYed up a bit (in fact, [rails has a mechanism for doing this](https://api.rubyonrails.org/classes/ActiveSupport/Concern.html)).

I deliberately didn't do this. DRY, like many principles, can actually be harmful if done across the board IMO. What I've chosen to do is keep all of the relevant logic in one place to make the code easier to read/follow and debug. Plus, if we end up adding additional rules in the future, we won't have to un-DRY things.

app/controllers/api/v1/comments/upvotes_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+              handle_upvote(vote)
+            end
+          end
+        end
+
+        private
+
+        def handle_existing_vote(vote)
+          if vote.upvote?
+            vote.destroy!
```
hernanat, line 19: the idea here is that if you click the upvote / downvote button twice, you want to un-do your vote. I could have had a separate endpoint for un-voting, but then we would have to introduce additional client-side logic (i.e. javascript) to know when to send a request to which endpoint. I despise having to do anything in javascript so I don't want to do this :)

hernanat:
[As a user I can upvote a submission](https://trello.com/c/w9pQI5lQ/4-as-a-user-i-can-upvote-a-submission)

hernanat:
[As a user I can upvote a comment](https://trello.com/c/FiwnyH46/5-as-a-user-i-can-upvote-a-comment)

hernanat:
[As a user I can downvote a submission](https://trello.com/c/DD0nQfG3/6-as-a-user-i-can-downvote-a-submission)

hernanat:
[As a user I can downvote a comment](https://trello.com/c/Pb08HQGH/7-as-a-user-i-can-downvote-a-comment)

hernanat:
thanks @ashpil !

app/controllers/api/v1/comments/downvotes_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+            if comment.descendants.where(user_id: current_user.id).exists?
```
hernanat, line 7: i knew this LTREE stuff I implemented would come in handy

hernanat: @ashpil it's worth noting here that the precondition is a more liberal one -- I'm not requiring the comment to be a _direct_ response to the downvoted comment, but to be among the comment's descendants. Do you think that this is a reasonable approach?

app/controllers/api/v1/comments/downvotes_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+              handle_vote(vote)
+            else
+              render json: build_response(:rejected).
+                merge(info: t(".comment_before_downvote")),
```
hernanat, line 11: the `info` key in the response body can be used to display a helpful message to the user (i.e. that they need to comment before downvoting)

app/controllers/api/v1/comments/downvotes_controller.rb
ashpil: I'm not sure if that makes sense to me. If there's a long tree of comments and a user contributes at the very bottom, they could be talking about something very different than what the original comment entailed. And if you comment at the very bottom you could get the ability to downvote tens of comments, rather than just one.

What do you see as the benefit of this approach?

hernanat: now that you say that, I realize that what I had in mind makes no sense. I'll fix that now. Thanks!

hernanat: THIS IS WHY HAVING COLLABORATORS IS AWESOME

hernanat: do you think the same argument applies to submissions? i.e. should they only be able to downvote if they have made a toplevel comment?

app/controllers/api/v1/comments/downvotes_controller.rb
```diff
@@ -0,0 +1,60 @@
+module Api
+  module V1
+    module Comments
+      class DownvotesController < BaseController
+        def update
+          find_or_initialize_vote.then do |vote|
+            if comment.children.where(user_id: current_user.id).exists?
```
hernanat, line 7: changed to require that the user directly responds to the comment being downvoted.

app/controllers/api/v1/comments/downvotes_controller.rb
ashpil: I think so, why not?

app/controllers/api/v1/submissions/downvotes_controller.rb
```diff
@@ -0,0 +1,60 @@
+module Api
+  module V1
+    module Submissions
+      class DownvotesController < BaseController
+        def update
+          find_or_initialize_vote.then do |vote|
+            if current_user.comments.where(submission_id: submission.id, parent_id: nil).exists?
```
hernanat, line 7: updated to require a top-level comment response for downvoting.

this is actually useful for helping save space in the votes table as well. though I wonder if we'll also be sort of artificially propping up the scores of some submissions given that some people might not want to be bothered to make a top level comment in order to downvote. hmm. we'll see what the feedback is in practice.

hernanat:
sorry @ashpil going to need another approval (unless you have anything else from my latest comments) since I pushed after you approved 😇 sorry ~

PR #37, opened 2020-05-22
make "Your Submissions" link work

hernanat:
https://trello.com/c/wa9VU6rA
https://trello.com/c/JBUhDJp3

this commit allows users to view their own submissions,
  and a side effect is that they can also view the submissions
  any other user by following the correct path.

When we build out the profile pages for users we'll have actual
  links to the latter.

### DIFF
--- a/app/views/layouts/_navbar.html.haml
+++ b/app/views/layouts/_navbar.html.haml
@@ -11,11 +11,13 @@
       %ul
         %li
           %span= link_to t("navigation.links.recent"), "#"
-        %li
-          %span= link_to t("navigation.links.submit"), new_submission_path
-        %li
-          %span= link_to t("navigation.links.your_submissions"), "#"
-        %li
-          %span= link_to t("navigation.links.inbox"), "#"
+        - if current_user.present?
+          %li
+            %span= link_to t("navigation.links.submit"), new_submission_path
+          %li
+            %span= link_to t("navigation.links.your_submissions"),
+              user_submissions_path(current_user.username)
+          %li
+            %span= link_to t("navigation.links.inbox"), "#"
         %li
           %span= link_to(*users_link_for(current_user))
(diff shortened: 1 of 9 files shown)


hernanat:
[as a user I can view my own submissions](https://trello.com/c/wa9VU6rA/38-as-a-user-i-can-view-my-own-submissions)

hernanat:
[as a user I can view the submissions by another user](https://trello.com/c/JBUhDJp3/39-as-a-user-i-can-view-the-submissions-by-another-user)

config/routes.rb
```diff
@@ -16,6 +16,8 @@
     resources :submissions, only: [:index]
   end
 
+  get "u/:username/submissions", to: "submissions#index", as: :user_submissions
```
hernanat, line 19: a side of this route, and the tags/:tag_id/submissions route, is that the url query parameters `:username` and `:tag_id` work in both cases.

i.e. if you want to get all of the submissions of a given user with a particular tag, you would go to:

`/u/theusername/submissions?tag_id=tag-id`

if you wanted to go the opposite way, and get all of the submissions under a given tag made by a particular user, it would be:

`/t/tag-id/submissions?username=theusername`

app/views/layouts/_navbar.html.haml
```diff
@@ -11,11 +11,13 @@
       %ul
         %li
           %span= link_to t("navigation.links.recent"), "#"
-        %li
-          %span= link_to t("navigation.links.submit"), new_submission_path
-        %li
-          %span= link_to t("navigation.links.your_submissions"), "#"
-        %li
-          %span= link_to t("navigation.links.inbox"), "#"
+        - if current_user.present?
```
hernanat, line 14: I'm hiding links that require a user context if their isn't a user signed in

hernanat:
thanks @ashpil

PR #40, opened 2020-05-22
adds votable score information to flattened database views

hernanat:
this commit adds the score to flattened submissions / comments,
  which is the difference between the number of upvotes and downvotes.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_05_21_044543) do
+ActiveRecord::Schema.define(version: 2020_05_22_214242) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "ltree"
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -145,10 +145,14 @@
       users.username AS submitter_username,
       submissions.original_author,
       submissions.created_at,
-      COALESCE(( SELECT count(comments.submission_id) AS comment_count
+      ( SELECT count(comments.submission_id) AS comment_count
              FROM comments
-            WHERE (comments.submission_id = submissions.id)
-            GROUP BY comments.submission_id), (0)::bigint) AS comment_count
+            WHERE (comments.submission_id = submissions.id)) AS comment_count,
+      (( SELECT count(votes.votable_id) AS count
```
hernanat, line 151: eventually it may or may not be beneficial to cache the scores on the submissions / comments records themselves, but then we have to remember to maintain that column so I'm avoiding it for now.

db/schema.rb
```diff
@@ -145,10 +145,14 @@
       users.username AS submitter_username,
       submissions.original_author,
       submissions.created_at,
-      COALESCE(( SELECT count(comments.submission_id) AS comment_count
+      ( SELECT count(comments.submission_id) AS comment_count
```
hernanat, line 148: I'm not sure why I was using coalesce / grouping previously (the coalesce was necessitated by the group by since the result would be `NULL` if there were no comments for a submission)

hernanat:
@ashpil this implements the backend for retrieving the score. You can call it in the views like any method call now -- i.e. for example in `app/views/submissions/_submission_list_item.html.haml` you would just call `submission.score`. Is that what you needed?

db/schema.rb
hernanat: it's worth noting that if we do ever get there, we can use this view query to backfill all scores for existing votable things, since there is a one-to-one correspondence between each "votable" record type and its flattened representation.

hernanat:
thanks ! also it appears github is having some issues right now :(

ashpil:
> 
> 
> @ashpil this implements the backend for retrieving the score. You can call it in the views like any method call now -- i.e. for example in `app/views/submissions/_submission_list_item.html.haml` you would just call `submission.score`. Is that what you needed?

And how would one do that for a comment?

hernanat:
flattened comments have the same method

ashpil:
Sorry, what is the difference between flattened and normal comments? 

Just adding `<span><% comment.score %></span>` to `app\views\comments\_comment_tree.html.erb` doesn't seem to do the trick.

ashpil:
Oh whoops that `=` was important, I got it now.

hernanat:
did you pull the latest master and run `bundle exec rake db:migrate` ?

A comment is connected to several different things -- a submission, a user,etc.

A flattened comment is a representation of the comment with all of the different things we care about from the comment and its related entities, squished into one thing. For example, we don't need the entire user record to show who authored the comment, just the username. So the flattened comment has all the comment information, plus the username (and other things).

Similarly, a comment's score is the number of upvotes minus the number of downvotes. If we wanted to do this in a normalized way, we load all of the comment's vote records into memory, and the count each kind and take the difference, for example. However, we don't use the vote records for anything else, so loading them into memory is a waste and instead we do the calculation in the database and return the result as part of the flattened comment as the `score`.

Does that make sense?

ashpil:
Yup, thanks!

PR #41, opened 2020-05-23
added/cleaned up css for voting/scoring

ashpil:
This isn't completely done yet, but I figured I'd do a PR because I needed some thoughts.

This is what the comment tree looks like now:

image

The vote buttons are capital deltas, which I sort of like thematically, though I'm not sure how well it works stylistically. Better than a plus/minus, I guess.

I haven't touched the post scoring yet. How are we going to implement the voting for them? Same way as the comments? It would be easier to implement all of the downvote/upvote/score styling at once, rather than just the score now and vote buttons later.

Also, now that the score is shown it's sort of unsatisfying to see you upvote/downvote and the score not change. What would be the best way to fix this? Is it worth it updating the database then requesting the new score, or, presumably, it would be better to just client side javascript change the score by one?

### DIFF
--- a/app/assets/stylesheets/shared/_votes.scss
+++ b/app/assets/stylesheets/shared/_votes.scss
@@ -1,0 +1,28 @@
+.scoring {
+    text-align: center;
+    .upvote,
+    .downvote {
+        cursor: pointer;
+        color: $scorpion;
+        line-height: 0.9;
+    }
+    .upvote::before {
+        content: "Δ";
+    }
+    .downvote::before {
+        content: "Δ";
+    }
+    .downvote {
+        -moz-transform: scale(1, -1);
+        -webkit-transform: scale(1, -1);
+        -o-transform: scale(1, -1);
+        -ms-transform: scale(1, -1);
+        transform: scale(1, -1);
+    }
+    .upvoted {
+        color: $orange;
+    }
+    .downvoted {
+        color: $purple;
+    }
+}
--- a/app/javascript/packs/comment.js
+++ b/app/javascript/packs/comment.js
@@ -30,24 +30,20 @@ pageload.onPageLoad(function() {
   for(let i = 0; i < hide_button_controls.length; i++) {
     let hide_button = hide_button_controls[i];
     hide_button.addEventListener("click", function(e) {
-      let commentId = hide_button.closest(".comment").id;
-      let commentTree = document.getElementById("tree_" + commentId);
-      let replies = document.getElementById("replies_" + commentId);
-      let upvote = document.getElementById("upvote_" + commentId);
-      let downvote = document.getElementById("downvote_" + commentId);
-      if (commentTree.style.display === "none") {
+      let comment = hide_button.closest(".comment");
+      let content = document.getElementById("comment_content_" + comment.id);
+      let scoring = document.getElementById("scoring_" + comment.id);
+      if (content.style.display === "none") {
         hide_button.classList.remove("hidden");
-        commentTree.style.display = "initial";
-        replies.style.display = "initial";
-        upvote.style.display = "initial";
-        downvote.style.display = "initial";
+        content.style.removeProperty("display");
+        scoring.style.removeProperty("display");
+        comment.style.removeProperty("grid-row-gap");
 
       } else {
         hide_button.classList.add("hidden");
-        commentTree.style.display = "none";
-        replies.style.display = "none";
-        upvote.style.display = "none";
-        downvote.style.display = "none";
+        content.style.display = "none";
+        scoring.style.display = "none";
+        comment.style.gridRowGap = "0";
       }
     });
   }
--- a/app/javascript/packs/voting.js
+++ b/app/javascript/packs/voting.js
@@ -11,18 +11,31 @@ function getOppositeControl(shortId, voteControl) {
 }
 
 function handleRemovedVote(votableElement, voteControl) {
+  let score = document.getElementById("score_" + votableElement.id);
   if (voteControl.id.startsWith("upvote_")) {
+    if (voteControl.classList.contains("upvoted")) {
+      score.innerHTML = Number(score.innerHTML) - 1;
+    }
     voteControl.classList.remove("upvoted");
-  } else if (voteControl.id.startsWith("downvote_")){
+  } else if (voteControl.id.startsWith("downvote_")) {
+    if (voteControl.classList.contains("downvoted")) {
+      score.innerHTML = Number(score.innerHTML) + 1;
+    }
     voteControl.classList.remove("downvoted");
   }
 }
 
 function handleUpdate(votableElement, voteControl, json) {
+  let score = document.getElementById("score_" + votableElement.id);
   if (json.action === "removed") {
     handleRemovedVote(votableElement, voteControl);
   } else {
     voteControl.classList.add(json.action);
+    if (json.action === "upvoted") {
+      score.innerHTML = Number(score.innerHTML) + 1;
+    } else {
+      score.innerHTML = Number(score.innerHTML) - 1;
+    }
     handleRemovedVote(votableElement, getOppositeControl(votableElement.id, voteControl));
   }
 }
@@ -60,12 +73,12 @@ function setVoteListener(votableElement, dir) {
     750,
     { 'trailing': false }
   );
-  voteControl.addEventListener("click", function(e) {
-    throttledRequest(votableElement, path, voteControl)
+  voteControl.addEventListener("click", function (e) {
+    throttledRequest(votableElement, path, voteControl);
   });
 }
 
-pageload.onPageLoad(function() {
+pageload.onPageLoad(function () {
   const votableElements = document.getElementsByClassName("votable");
 
   for (let i = 0; i < votableElements.length; i++) {
--- a/app/views/submissions/_submission_list_item.html.haml
+++ b/app/views/submissions/_submission_list_item.html.haml
@@ -1,22 +1,28 @@
-.submission{ id: submission.short_id, class: submission_classes_for(submission) }
-  .submission-line1
-    %span.submission-link= link_to(submission.title, submission_href_for(submission))
-    %span.submission-tags
-      - submission.tags.each do |tag|
-        = render "shared/tag", tag: tag
-    %span.submission-info
-      %i= submission.domain_name.presence
-  .submission-line2
-    %span.submission-info
-      = submitted_by_text(submission)
-    %span.submission-user
-      = link_to submission.submitter_username, "#"
-    %span.submission-info
-      %span.submission-time
-        = submitted_time_text(submission)
-    %span.submission-info
-      = link_to hide_link_text_for(submission), "", class: "hide"
-      = link_to save_link_text_for(submission), "", class: "save"
-    %span.submission-comments
-      = link_to t(".comments", count: submission.comment_count),
-        submission_path(submission.short_id)
+.submission.votable{ id: submission.short_id, class: submission_classes_for(submission) }
+  %span.scoring{ id: "scoring_" + submission.short_id }
+    .upvote{ id: "upvote_" + submission.short_id, class: control_classes(submission, :upvote) }
+    .score{ id: "score_" + submission.short_id, class: control_classes(submission, :score) }
+      = submission.score
+    .downvote{ id: "downvote_" + submission.short_id, class: control_classes(submission, :downvote) }
+  .submission-title
+    .submission-line1
+      %span.submission-link= link_to(submission.title, submission_href_for(submission))
+      %span.submission-tags
+        - submission.tags.each do |tag|
+          = render "shared/tag", tag: tag
+      %span.submission-domain
+        %i= submission.domain_name.presence
+    .submission-line2
+      %span.submission-info
+        = submitted_by_text(submission)
+      %span.submission-user
+        = link_to submission.submitter_username, "#"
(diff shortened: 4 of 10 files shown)


app/assets/stylesheets/shared/_votes.scss
```diff
@@ -0,0 +1,19 @@
+.upvote,
```
ashpil, line 1: Honestly, not a lot to do here. Most of the placement is done in the submission CSS file, because that's specific to the comment. Now that I'm thinking about it, this might be cleaner if I redo it without grid. Might just do that.

app/javascript/packs/comment.js
```diff
@@ -30,24 +30,31 @@ pageload.onPageLoad(function() {
...
-        replies.style.display = "initial";
-        upvote.style.display = "initial";
-        downvote.style.display = "initial";
+        commentTree.style.removeProperty("display");
+        replies.style.removeProperty("display");
+        upvote.style.removeProperty("display");
+        downvote.style.removeProperty("display");
+        score.style.removeProperty("display");
+        comment.style.removeProperty("grid-template-rows");
+        comment.style.removeProperty("grid-row-gap");
```
ashpil, line 47: I had to change this a bit due to grid, as this makes it actually collapse properly. I think it's cleaner this way too.

hernanat:
> How are we going to implement the voting for them? Same way as the comments?

I'm not sure what you mean -- the backend part is already implemented. Do you mean where will they go in the UI? I think it makes sense to have them follow the same format as the comments.

> Also, now that the score is shown it's sort of unsatisfying to see you upvote/downvote and the score not change.

We can add the new score as part of the API response and update it on the client side given that information. Or we can manually change the count on the client side which might make more sense. The nature of these things is that someone might be upvoting / downvoting at the same time and so whatever score the server returns might actually be out of sync as soon as you get it, so if the point is aesthetics I think it makes more sense to just increment / decrement the score count using javascript on the client side rather than make additional work for the server.

hernanat:
> The vote buttons are capital deltas

technically the downvote is a nabla :P

app/assets/stylesheets/shared/_votes.scss
```diff
@@ -0,0 +1,19 @@
...
+.downvote {
+    cursor: pointer;
+    color: $scorpion;
+    line-height: 0.9;
+    font-family: KaTeX_Main, $font-stack;
```
hernanat, line 6: this might cause errors in production environments, we'll have to double check

app/javascript/packs/comment.js
hernanat: I'll take your word for it. I'm about as good of a javascript programmer as I am a doctor. That is to say not at all.

If this ends up causing issues later we can revisit it.

My only concern with our javascript assets are their growing size; I also wonder what the performance effect is of having to do a ton of sequential modifications to the dom for each action.

Part of me was wondering if we could group things under a common class or something and collapse them that way. We can iterate 🤷

hernanat:
I assume downvotes are still purple?

ashpil:
> 
> 
> I assume downvotes are still purple?

Yup, that seemed fine to me.

app/assets/stylesheets/shared/_votes.scss
ashpil: Yeah, seems like something that might. Seemed to work just fine for me, though.

ashpil:
> I'm not sure what you mean -- the backend part is already implemented. 

I see. I just wasn't sure what you used to make the buttons actually do something. Can you tell me where that happens in the code for the comments and I can base the submissions version off of that?

app/javascript/packs/comment.js
ashpil: > Part of me was wondering if we could group things under a common class or something and collapse them that way.

I'll try this.

app/assets/stylesheets/shared/_votes.scss
hernanat: if you run `RAILS_ENV=production bundle exec rake assets:precompile` and open your browser sometimes that will tell right away whether or not something is broken (or the assets will just fail to compile.

hernanat:
ah I see. If you look at `app/javascript/packs/voting.js`, I wrote it with the intention of having it easily hook into voting for submissions. The initializing lines at the bottom are what you need. Basically it assumes that something is "votable" if it has the `votable` class, and from there it tries to find the controls, and will automatically set the correct listener based on the short id prefix (`s_` is for submission, `c_` is for comment).

ashpil:
Making progress...

image

Could you give me a hint as to how to make your previous voting visible on page load?

hernanat:
> Could you give me a hint as to how to make your previous voting visible on page load?

what do you mean? like to assign upvoted / downvoted classes to submissions that you've upvoted / downvoted?

app/assets/stylesheets/shared/_votes.scss
```diff
@@ -1,19 +1,22 @@
...
-    color: $purple;
-}
+.scoring {
+    text-align: center;
+    .upvote,
+    .downvote {
+        cursor: pointer;
+        color: $scorpion;
+        line-height: 0.9;
+        font-family: KaTeX_Main, $font-stack;
```
ashpil, line 8: I found the downfall of this: 

image

The KaTeX font doesn't load on the main page, so the nabla isn't properly shown there. Not sure how to fix this - I suppose we could just replace 'em with svgs, but I'm not too eager to do that as I feel like we're already loading it.

app/javascript/packs/comment.js
ashpil: I was able to simplify these quite nicely in another commit, though I didn't fully implement a common class.

app/views/submissions/_submission_list_item.html.haml
```diff
@@ -1,22 +1,28 @@
...
-      = link_to save_link_text_for(submission), "", class: "save"
-    %span.submission-comments
-      = link_to t(".comments", count: submission.comment_count),
-        submission_path(submission.short_id)
+.submission.votable{ id: submission.short_id, class: submission_classes_for(submission) }
+  %span.scoring{ id: "scoring_" + submission.short_id }
+    .upvote{ id: "upvote_" + submission.short_id }
+    .score{ id: "score_" + submission.short_id }
+      = submission.score
+    .downvote{ id: "downvote_" + submission.short_id }
```
ashpil, line 6: Could you make sure this actually looks like it's supposed to, @hernanat, I'm not sure what you're expecting here.

ashpil:
> 
> 
> > Could you give me a hint as to how to make your previous voting visible on page load?
> 
> what do you mean? like to assign upvoted / downvoted classes to submissions that you've upvoted / downvoted?

Yeah

hernanat:
if you look at `app/models/flattened_comment.rb`, there is a class method called

`with_voting_information_for` -- basically you need to do that for `FlattenedSubmission`. You can take the same code and in the SQL where it says :

```sql
...
votes.votable_type = 'Comment'
...
```
you want to replace `'Comment'` with `'Submission'`. Similarly where you see `flattened_comments`, replace it with `flattened_submissions`.

Then in `FlattenedSubmissionSearch`, you can change the `base_relation` method to this:

```ruby
  def base_relation(include_hidden: false)
    FlattenedSubmission.preload(:tags).then do |rel|
      if user.present?
        with_action_info(rel, include_hidden)
      else
        rel.select("flattened_submissions.*, NULL AS hidden_action_id, NULL AS saved_action_id")
      end
    end.with_voting_information_for(user)
  end
```

this should do what you want. If you have trouble lmk and I'll do it. Right now I'm trying to work on a Dockerfile for our app

ashpil:
> Dockerfile 

That sounds useful, Docker is one of those things that I hear about everywhere but haven't actually put in the effort into understanding what it is yet.

hernanat:
> That sounds useful, Docker is one of those things that I hear about everywhere but haven't actually put in the effort into understanding what it is yet.

hernanat:
> That sounds useful, Docker is one of those things that I hear about everywhere but haven't actually put in the effort into understanding what it is yet.

Super useful, and one of those things that I like to nerd out about. A hobby horse of mine is trying to get the resulting containers as small as possible 😁 I'm using Alpine linux for ours.

ashpil:
Aight, that should do it.

app/javascript/packs/voting.js
```diff
@@ -11,18 +11,31 @@ function getOppositeControl(shortId, voteControl) {
...
+  let score = document.getElementById("score_" + votableElement.id);
   if (json.action === "removed") {
     handleRemovedVote(votableElement, voteControl);
   } else {
     voteControl.classList.add(json.action);
+    if (json.action === "upvoted") {
+      score.innerHTML = Number(score.innerHTML) + 1;
+    } else {
+      score.innerHTML = Number(score.innerHTML) - 1;
+    }
```
ashpil, line 38: Please make sure my javascript here makes sense, I think my JS is even worse than yours. Also I think my auto-formatter added spaces between all functions and the `()`, weird.

app/assets/stylesheets/shared/_votes.scss
hernanat: that's because I currently have it setup so that the katex library only gets loaded on pages where there is latex being displayed. We would have to make that font available across the application which is enough enough to do.

ashpil: What do you think we should do? Svgs? Load katex on main? Scrap the deltas/nablas completely?

hernanat: it does seem like overkill to load the entire font for just the upvote controls. Plus it gets loaded on the pages that have LaTeX anyways which means we might be loading it twice? (I'm not sure if Rails manages assets intelligently enough to avoid this).

I'm not sure how hard it would be to extract just those characters either. I would have to think / explore this a bit to get any ideas for what a good solution might be.

In the interest of saving time and keeping the work you've done, what do you think about just loading the font for the entire app?

it's located in `node_modules/katex/dist/fonts`. We can come back to this when we have a better idea of what to do.

Also I think there is a way to do this with webpacker which might take care of not loading the same font twice, but that's a rabbit hole I don't really want to go down this late at night.

ashpil: Actually.... Might just be able to use the Delta and flip it upside-down with css. Going to try that.

ashpil: Yup, that seems to work great!

hernanat:
nice commit message 😂

ashpil:
One issue that is left is that for some reason you don't seem to be able to vote on submissions from the main page. Any idea why that might be happening, @hernanat? I assume it's because the voting script isn't loaded on the main page. How would I load it there as well?

hernanat:
> One issue that is left is that for some reason you don't seem to be able to vote on submissions from the main page. Any idea why that might be happening, @hernanat? I assume it's because the voting script isn't loaded on the main page. How would I load it there as well?

add in

```haml
= javascript_pack_tag "voting"
```

at the bottom of `app/views/submissions/index.html.haml`

I avoided including all the javascript everywhere to prevent slowing down pages that don't need it, but the downside is that it's not clear why things don't just "work" 😇 

I've been trying to think of a better way to manage this.

ashpil:
> I avoided including all the javascript everywhere to prevent slowing down pages that don't need it, but the downside is that it's not clear why things don't just "work" 😇

I'm totally with you, less javascript == best javascript.

ashpil:
This look good for a squash + merge now?

hernanat:
yeah looks good. We should probably add tests for this but I trust that it works and we can always add them in tomorrow. Let's get this merged + deployed so we can test it in a prod-ish environment.

ashpil:
> 
> 
> yeah looks good. We should probably add tests for this but I trust that it works and we can always add them in tomorrow. Let's get this merged + deployed so we can test it in a prod-ish environment.

Sweet. Honestly, test are another one of those things I know absolutely nothing about. I figured you only needed them for backend/ruby changes?

hernanat:
> Sweet. Honestly, test are another one of those things I know absolutely nothing about. I figured you only needed them for backend/ruby changes?

hernanat:
> Sweet. Honestly, test are another one of those things I know absolutely nothing about. I figured you only needed them for backend/ruby changes?

I'm of the opinion that every piece of code / functionality that can be feasibly tested should be. Otherwise there isn't a real way of knowing if a change in one place breaks something in another place for example. You might have noticed that many of my PRs contain more tests than actual application code -- I think this has a few benefits:

- as I mentioned, if there are unintentional side effects of your changes, chances are they will break a test (if you've tested well). At this point you can figure out whether your change isn't entirely correct, or you need to change that other code as well.

- demonstrating that your code "works" -- if you write good tests and they all pass, chances are you're not going to blow other things up

- making sure individual pieces of code work vs trying to look at everything altogether all the time. As applications grow, they become more and more complex, and so doing simple smoke tests on new changes becomes less reliable.

In rails / our app (and other frameworks as well) we also have what are called system specs which are used to test how the user interacts with the system as a whole and that the expected behavior actually occurs. These are typically what we write when we want to test front-end related things (check out `spec/system`).

It's definitely a useful thing to know. I've found that testing also helps me feel a bit more confident about the code that I ship as well.

Now obviously no one is perfect and software is prone to issues, so testing won't save you from _everything_, but I think it saves you from many things.

And now that I've gone on and on about how much I love testing, we're probably going to ship some insane bug in the near future 😂

PR #42, opened 2020-05-25
adds configurations for mailing

hernanat:
We need to be able to mail stuff to folks for
  things like email confirmations.

This enables us to do that.

I've set us up with Amazon SES for use in production.

At the moment I just have it using my own domain, but once
  we have an actual domain I'll configure it for that.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,2 +1,6 @@
 # raise exceptions when I18n translations are missing; useful for development + testing
 RAISE_ON_MISSING_TRANSLATIONS=TRUE
+SMTP_SENDER_EMAIL=foo@bar.com
+SMTP_ADDRESS=smtp.domain.com
+SMTP_USERNAME=foop
+SMTP_PASSWORD=foobar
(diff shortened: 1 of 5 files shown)


hernanat:
Also, but at what cost:

Screen Shot 2020-05-25 at 1 35 07 PM


I guess there's a perk to hosting with amazon

hernanat:
> By the way, what still needs to be done before a docker PR?

Not much really but what I want to tear down all of the environments and recreate them from scratch so I can document the process / verify that it's reproducible. I also want to maybe write a script to help with some of the stuff.

There are two sides to it -- the `docker-compose` side which I think will be useful for development, and the `eb` CLI side, which uses `docker-compose` under the hood but configures things in a slightly different manner. I want to make sure both of these work at the local level as expected so that folks can choose to use either / both without being tied down.

I know you're eager to try it out so I'll try to do it today :)

hernanat:
thanks @ashpil; I'll try to deploy this later on so we can try it in a production environment. Not able to at the moment.

PR #43, opened 2020-05-26
Docker

hernanat:


hernanat:
this isn't quite done yet

config/credentials.yml.enc
```diff
@@ -1 +1 @@
-egGwa5K2iBCM1wGtf6Fi4094J7YDxjkfpqS5YQjBD5FekfhD0WKwHEpC2mPhOGEaqin86ETVBd5JUdZ4AAKOkRYtp9R1uSrymrBGEvA2UMmF8kr/2vHyofoRkUVpwc9LtZhLWAxuJ7WN991WNQkGD/v7OFUpObarH+wh53Nf3P9rSmSJc66aFhEZx3Rrzfkf2E+dYSnLpzAXSeRE3g6a4TnHm6kENQPnC0gdhVI1smEjpsvVoxj3Z2SJ+Ez+dM6AvyUZwrIRLXxtK0cmQdArgQf2Uq12RISdPDPX5l4T3Ec/fT1jNDFEOVdrNksCTE3exWlW0wo6yUE3E7z8Go3jgVBGGfGLWwCQCjUKN7L81of77ofSy6SahVUVSa2vninvfr5g67rMp+eRamVIOjIl2uVqlsfQaBfoS9mM--eiM062Nvdv/Iw2hV--kcoKbQNyZEZ7S6jBS8hWzw==
\ No newline at end of file
+Tv6+kBN5yNWJmFGgRJ6AJi7YCMsZRlIdza/Ru1ZCebMkKvCMBUEies6Oswr0QBKc3ruFYzSiEfnQ+Ll1rggIcJHu5oUfcELm7x9z4mQNso56dnoOn3OO+5+0UlB/ePREDXG7tMVuhV/DNkiMfdDhN9tF6tX8km3r1AUWaCQOaFgcBESApfRIvugAVB111dU6NO6Oqw666pYailf7ky6tBCzJpqsFzfCGsLk8VTJRPSn72evraQD5Kob74X/vWsVSBopH8cIO7wsUMiVKxMA2iukIM8j7Sb4ygC+JGr185AmA61TxCIsIgBStoaWMW4R1sQniUFdJGeeFOQkWnJ6sAJ4ii9NmOTQx8Qh7kuk05d47WN8vrzgLcNt9bUxOc4vkJkymCTMmCxTrPGpTPhoUcNHAMxCY6jJsZLdX--cAjqa0BvHZCD2Nwp--PdTduhjk1+ne2RNenOo32g==
```
hernanat, line 1: I accidentally lost the other master key file used to encrypt these so I had to regenerate it. All it contains is the rails secret key. I'm not sure it's even really keeping this file around though since      we're just using environment variables anyways

config/database.yml
```diff
@@ -3,6 +3,9 @@ default: &default
   encoding: unicode
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   url: <%= ENV.fetch("DATABASE_URL", "") %>
+  host: <%= ENV.fetch("DATABASE_HOST", "") %>
```
hernanat, line 6: I needed to add this in for when folks are running the app inside of a docker environment in development.

config/environments/development.rb
```diff
@@ -61,4 +61,6 @@
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+
+  config.webpacker.check_yarn_integrity = false
```
hernanat, line 65: I think I'm going to put this behind an environment variable given that this only becomes an issue when you're developing within the dev docker context. basically without this it always thinks your yarn packages are out of date and won't start the server.

docker-compose.development.yml
```diff
@@ -0,0 +1,26 @@
+version: '3'
+services:
+  app:
+    image: hernanat/simple-group-app-dev:latest
```
hernanat, line 4: I gave a development version of this in case anyone decided that they wanted to use it. It creates a container running alpine linux that you can run the app / tests in if you wish, and it also creates a postgres database in a separate container also running on alpine linux. I like alpine because it is v smol.

If you wanted to run this setup, you would do this:

```
docker-compose -f docker-compose.development.yml --env-file .env up --build
```

And if you wanted to get into the app container:

```
docker exec -it simple_group_app_1 /bin/sh
```

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+services:
+  app:
+    image: hernanat/simple-group-app-prod:latest
+    ports:
+      - "3000:3000"
+    build:
+      context: .
+      dockerfile: ./docker/app/Dockerfile.production
+      args:
+        - APPLICATION_HOST=placeholder
```
hernanat, line 11: unfortunately we have some environment variables that are required tangentially while we're running various commands at build time (i.e. precompile assets, etc.).

A docker `ARG` is a variable passed to the image at build time but does not persist after the build has finished. It can be accessed within the build context just like an environment variable.

Contrast that with docker `ENV`s which set environment variables that _persist into the image itself_. This is not something that we want for things like secrets / credentials, especially since this application is in the public domain.

The `environment` declarations below are responsible for passing environment variables to the container at run time.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: '3'
+services:
```
hernanat, line 2: if you're running this locally you're going to run into issues when it comes to connecting to a database. By default it's not entirely straightforward to connect to a database running on the host machine. You can either:

1.) do some hacky postgres configurations to allow the docker container to connect to the database through the host (your computer) ip address
2.) use an actual database URL like you would be doing in production.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+      - APPLICATION_HOST=${APPLICATION_HOST}
+      - RAILS_ENV=production
+      - SECRET_KEY_BASE=${SECRET_KEY_BASE}
+      - SMTP_SENDER_EMAIL=${SMTP_SENDER_EMAIL}
+      - SMTP_ADDRESS=${SMTP_ADDRESS}
+      - SMTP_USERNAME=${SMTP_USERNAME}
+      - SMTP_PASSWORD=${SMTP_PASSWORD}
+      - DATABASE_URL=${DATABASE_URL}
+
+  nginx:
```
hernanat, line 28: nginx is being used to serve our precompiled static assets in production environments so that the rails server doesn't have to.

the way that I have this setup in is such that `nginx` listens on port `80` for all incoming http traffic. If the request is for a static asset, then nginx will serve it. Otherwise, it forwards the request onto the rails server. More on this later.

AWS / other hosting providers typically have an external service listening on port 80/443 for http/https traffic, and then forward it to your application port. In our case, we have an elastic load balancer that I've configured to do the following:

- listen on port 443 (https) and forward the incoming traffic to port 80 on the nginx container
- listen on port 80 (http) and redirect the incoming traffic to port 443 (this forces SSL)

That we're able to do this is nice because it avoids making me have to do shit relating to SSL certificates with nginx manually since SSL is taken care of at the load balancing layer (I created an SSL certificate for my domain to use on AWS and configured the load balancer listener on port 443 with it)

This setup is very similar to the one that is used in production, but it's not exactly the one. One reason for this is that beanstalk generates its own `docker-compose` setup under the hood given a `Dockerrun.aws.json` file (I don't understand why they wouldn't just let us use a docker compose file directly because the json thing they have you do is exponentially more complicated imo). Another reason is that there are some beanstalk-specific directory related things in that setup that I didn't want to include here. I opted to leave the beanstalk setup out of this PR.

Both setups use the same dockerfiles though at least.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    image: hernanat/simple-group-nginx-prod:latest
+    restart: always
+    build:
+      context: ./docker/web
+      dockerfile: Dockerfile.production
+    volumes:
+      - ./log:/var/www/log
```
hernanat, line 35: a volume is a directory shared between the host machine (left) and the container. In our case we're sharing the log directory local to the app code. nginx will write its log files to this directory.

docker/app/Dockerfile.development
```diff
@@ -0,0 +1,42 @@
+FROM ruby:2.7.1-alpine3.11 as Builder
+
+ARG RAILS_ENV
```
hernanat, line 3: I don't know if I actually need these in the dev environment since we're not precompiling or trying to run anything at build time. I'll have to double check.

docker/app/Dockerfile.production
```diff
@@ -0,0 +1,52 @@
...
+	&& apk add --update --no-cache --virtual runtime-dependencies \
+		postgresql-client \
+		postgresql-dev \
+		tzdata \
+		yarn \
+	&& bundle config set deployment 'true' \
+	&& bundle config set without 'development test' \
+	&& bundle install \
+	&& apk del build-dependencies \
+	&& rm -rf /opt/rails/vendor/bundle/ruby/*/cache/*.gem \
```
hernanat, line 25: I'm removing all the build artifacts to make the resulting image v smol

docker/app/Dockerfile.production
```diff
@@ -0,0 +1,52 @@
+FROM ruby:2.7.1-alpine3.11 as Builder
```
hernanat, line 1: I'm declaring this `as` something for some magic later

docker/app/Dockerfile.production
```diff
@@ -0,0 +1,52 @@
...
+
+RUN yarn install
+
+ADD . /opt/rails
+
+RUN bundle exec rake assets:precompile \
+	&& rm -rf node_modules tmp/cache vendor/assets lib/assets \
+	&& find ./app/assets -mindepth 1 -name config -prune -o -exec rm -rf {} +
+
+FROM ruby:2.7.1-alpine3.11
```
hernanat, line 39: this is where the magic happens. In the steps above, we had to install a bunch of stuff in order to build the things necessary to run our application (i.e. a `C` compiler to compile nokogiri, a dev library to install the `pg` gem, etc.).

We don't actually need those at run time, so there is no use in having them take up space in our resulting image.

This second `FROM ruby:2.7.1-alpine3.11` _wipes the slate clean for the image_. i.e. at this point it knows nothing about everything we did above. We then install some needed runtime dependencies, and the `COPY` only the things (mostly -- there is still some cleanup I can probably do) that we need at runtime from above using the

```
COPY --from=Builder ...
```

syntax. This is one of my favorite features of Docker.

docker/web/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM hernanat/simple-group-app-prod:latest AS Builder
```
hernanat, line 1: so maybe this is a bit wonky but hear me out:

nginx needs to be able to access the static files to serve them. we could do this in a few ways:

- precompile them, place them on the ec2 instance, and then define a shared volume

- have it pull them directly from the app container

The first choice requires us to precompile the assets an additional time.

The second choice requires nginx to talk to the rails app, and I don't really think it ought to have to.

What I've done is have the nginx container image actually pull the static assets from the app prod image at build time so that it has them already. Since we have to rebuild our app container image every time we make code changes anyways, this is just another part of that.

hernanat: it's worth mentioning that we could also just have the app serve static assets itself, but this is just additional load on the server. I'm also planning to explore caching options with nginx to further improve performance relating to serving static assets.

docker/web/nginx.conf
```diff
@@ -0,0 +1,39 @@
...
+    root   /var/www/public;  
+    index  index.html;
+    access_log /var/www/log/nginx.access.log;  
+    error_log /var/www/log/nginx.error.log;   
+    location ~ /\. {    
+      deny all;  
+    } location ~* ^.+\.(rb|log)$ {    
+      deny all;  
+    }  
+    location ~ ^/(assets|packs)/ {
```
hernanat, line 18: this is where the static magic is

docker/web/nginx.conf
```diff
@@ -0,0 +1,39 @@
...
+    index  index.html;
+    access_log /var/www/log/nginx.access.log;  
+    error_log /var/www/log/nginx.error.log;   
+    location ~ /\. {    
+      deny all;  
+    } location ~* ^.+\.(rb|log)$ {    
+      deny all;  
+    }  
+    location ~ ^/(assets|packs)/ {    
+      try_files $uri @rails;
```
hernanat, line 19: basically if the uri matches an `assets` or `packs` request, we'll go in here. nginx will try to serve the request and if it can it will forward it to the rails app.

hernanat:
@ashpil I've tried to give some helpful explanations. This is still very much a work in progress but let me know if you have any questions. It's probably worth trying to get the development docker environment up and running. you'll need `docker` and `docker-compose`

hernanat:
here are the resulting image sizes:

```
REPOSITORY                         TAG                 IMAGE ID            CREATED             SIZE
hernanat/simple-group-nginx-prod   latest              d5630bcfe950        8 minutes ago       31.1MB
hernanat/simple-group-app-prod     latest              d5285830fcf9        9 minutes ago       142MB
hernanat/simple-group-app-dev      latest              213b9c9905f8        10 minutes ago      266MB
```

hernanat:
I'm off vacation tomorrow so I won't be able to do much work until after work, but we're really close to launch it seems like. Super stoked to get people using this. Now we just need to figure out the naming / domain situation, otherwise the app is just going to live on a subdomain of my domain forever haha (which maybe in the interim it's not a bad idea to have `pascal.antfeedr.com` and if we ever get the other domain I can just forward any new requests to subdomain to there).

ashpil:
Wow. This is amazing - and I admit I haven't seen many docker setups - but this is the most complicated one I've ever seen, lmao. 

Couple questions/remarks whose answers are likely obvious to those who have any sort of experience with docker:

1. What's the reasoning behind having rails/postgres/nginx in separate containers, rather than one?
2. Is really having all that extra stuff to compile nokogiri/etc and then wipe it clean the best way to do it? I would've figured this sort of stuff comes precompiled and essentially ready to use (or a way to get it that way).

I'll try to set this up tomorrow on my machine, see if it works :p (and cuz docker I assume it should!)

Then, if necessary, I'll also write a README for the Docker installation, make the process a bit more clear (unless of course you wanna do that). 

Btw, I've emailed the guy about the domain. Awaiting a response. I suppose we could go with one of the less popular options, but I'm not too keen on a rebranding later if pasc.al opens up.

hernanat:
oh the other thing ... elastic beanstalk doesn't allow building docker images at deployment for multicontainer docker environments for whatever reason...it literally forces you to have the image stored in a repository somewhere. So our deployments in this case are actually quite interesting -- we need to _rebuild_ the environments after we deploy to them.

but the funny thing is that we don't really need to deploy to them at all, because all we really need is for the environment to have the most recent docker images (which have the most recent code), and it only does that when you rebuild it. So basically whenever we want to "deploy", we have to push our docker images to a docker repo first, and then trigger a rebuild in elastic beanstalk. fkcing amazon.

i.e. we actually don't need to deploy any of our code directly to elastic beanstalk at all since it resides within the docker images (I actually want to make it so that we don't deploy our codebase, which will allow us to reduce our memory footprint further).

ashpil:
> I actually want to make it so that we don't deploy our codebase, which will allow us to reduce our memory footprint further.

It'll be interesting to see how the sizes of the `dev` and `prod` containers compare by the end of this.

hernanat:
> What's the reasoning behind having rails/postgres/nginx in separate containers, rather than one?

For one it's the "docker" way -- one process per container. But I think we can rationalize it ourselves.

Let's say that we want to scale something horizontally (i.e. add more containers). If we had kept everything on one container, our options are:

1.) add _everything_ to that container to more containers

2.) create a new image which contains the parts of the original image that we need to scale whatever service horizontally.

I think (1) is clearly not desirable, it's extra stuff that we don't need! So we would end up with (2), which is basically what I've done here.

Let's also consider the case where (for whatever reason) the container becomes compromised. In the case where all three services are living in one container, all three services have effectively been compromised. Contrast that with our situation. Let's suppose that the nginx container becomes compromised. Since the nginx and the rails container are separate entities, we've reduced our problem space to only having to address the compromised nginx container (we should of course make sure none of the other ones have been compromised).

There could also be the situation where certain software updates might break certain services and not others. For example, suppose that for whatever reason postgres doesn't behave nicely on the newest version of Alpine linux, but everything else runs fine. Suppose further that the latest version has updates that the other service(s) actually benefit greatly from. What do we do in this situation? If they all live in one container, the problem is much more difficult. However, if they all live in separate containers, we can just bump the postgres image's OS back down.

We can take this a step further -- what if we wanted to run Rails on Debian, Postgres on Ubuntu, and nginx on Alpine? Or maybe we have some service that runs exclusively on microsoft operating systems, but we don't want to run our other services on microsoft (if I had to run Rails on Windows I might actually quit).

> Is really having all that extra stuff to compile nokogiri/etc and then wipe it clean the best way to do it? I would've figured this sort of stuff comes precompiled and essentially ready to use (or a way to get it that way).

Do you mean are there binary packages? That depends. Nokogiri, for example, only officially maintains three binary packages: https://nokogiri.org/#binary-packages

Similarly, other projects may only maintain a small number of binary package distributions, and they might not even work for the OS you're running your services on. This is especially true in the case of Alpine which we're using -- it uses the musl C library whereas most other distributions use glibc.

It's true that we need to install several build packages to get up and going but it doesn't take too long. Alpine ships with the bare minimum to run the operating system. The result is that you don't have all of the extraneous crap that you're never going to use taking up space on your app container. For other container images for things like debian / ubuntu, the same is not true. In our case, we know exactly what extras were installed and so we know exactly what we want to remove at the end. For other images, they come with a plethora of other things preloaded and so trying to cleanup those could be a huge headache.

ashpil:
Sweet, those answers make sense and are more or less what I expected. I'll try this out tomorrow and let you know how it goes.

hernanat:
I'm playing around with having nginx actually cache assets vs just storing them locally and the results are interesting. However results locally aren't really a good testament to what will happen in production so I'll have to setup testing environments for each scenario and see:

- the "control" being rails serving static assets

- nginx serving static assets w/o caching

- nginx serving static assets w/ caching (and at this rate we can probably just turn on static assets for the rails app and whenever we need to refresh the cache it will just go to the rails app itself. this seems like a much more efficient use of resources vs storing the assets in both containers; it will also simplify our nginx image)

hernanat:
I should probably take a break from this rabbit hole and work on some actual features thouh...hm

docker-compose.development.yml
ashpil: Welllllll, something went wrong.

I think.

Upon running that top command initially, first I get what looks all correct, then this:
```
Successfully built 5d7917b3f5c6
Successfully tagged hernanat/simple-group-app-dev:latest
Creating simple_group_db_1 ... done
Creating simple_group_app_1 ... done
Attaching to simple_group_db_1, simple_group_app_1
db_1   | The files belonging to this database system will be owned by user "postgres".
db_1   | This user must also own the server process.
db_1   | 
db_1   | The database cluster will be initialized with locale "en_US.utf8".
db_1   | The default database encoding has accordingly been set to "UTF8".
db_1   | The default text search configuration will be set to "english".
db_1   | 
db_1   | Data page checksums are disabled.
db_1   | 
db_1   | fixing permissions on existing directory /var/lib/postgresql/data ... ok
db_1   | creating subdirectories ... ok
db_1   | selecting dynamic shared memory implementation ... posix
db_1   | selecting default max_connections ... 100
db_1   | selecting default shared_buffers ... 128MB
db_1   | selecting default time zone ... UTC
db_1   | creating configuration files ... ok
db_1   | running bootstrap script ... ok
db_1   | performing post-bootstrap initialization ... sh: locale: not found
db_1   | 2020-05-26 22:53:05.616 UTC [30] WARNING:  no usable system locales were found
db_1   | ok
db_1   | syncing data to disk ... initdb: warning: enabling "trust" authentication for local connections
db_1   | You can change this by editing pg_hba.conf or using the option -A, or
db_1   | --auth-local and --auth-host, the next time you run initdb.
db_1   | ok
db_1   | 
db_1   | 
db_1   | Success. You can now start the database server using:
db_1   | 
db_1   |     pg_ctl -D /var/lib/postgresql/data -l logfile start
db_1   | 
db_1   | waiting for server to start....2020-05-26 22:53:05.945 UTC [35] LOG:  starting PostgreSQL 12.3 on x86_64-pc-linux-musl, compiled by gcc (Alpine 9.2.0) 9.2.0, 64-bit
db_1   | 2020-05-26 22:53:05.948 UTC [35] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db_1   | 2020-05-26 22:53:05.967 UTC [36] LOG:  database system was shut down at 2020-05-26 22:53:05 UTC
db_1   | 2020-05-26 22:53:05.971 UTC [35] LOG:  database system is ready to accept connections
db_1   |  done
db_1   | server started
db_1   | 
db_1   | /usr/local/bin/docker-entrypoint.sh: ignoring /docker-entrypoint-initdb.d/*
db_1   | 
db_1   | waiting for server to shut down....2020-05-26 22:53:06.038 UTC [35] LOG:  received fast shutdown request
db_1   | 2020-05-26 22:53:06.042 UTC [35] LOG:  aborting any active transactions
db_1   | 2020-05-26 22:53:06.042 UTC [35] LOG:  background worker "logical replication launcher" (PID 42) exited with exit code 1
db_1   | 2020-05-26 22:53:06.042 UTC [37] LOG:  shutting down
db_1   | 2020-05-26 22:53:06.065 UTC [35] LOG:  database system is shut down
app_1  | Puma starting in single mode...
app_1  | * Version 4.3.5 (ruby 2.7.1-p83), codename: Mysterious Traveller
app_1  | * Min threads: 5, max threads: 5
app_1  | * Environment: development
db_1   |  done
db_1   | server stopped
db_1   | 
db_1   | PostgreSQL init process complete; ready for start up.
db_1   | 
db_1   | 2020-05-26 22:53:06.146 UTC [1] LOG:  starting PostgreSQL 12.3 on x86_64-pc-linux-musl, compiled by gcc (Alpine 9.2.0) 9.2.0, 64-bit
db_1   | 2020-05-26 22:53:06.146 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
db_1   | 2020-05-26 22:53:06.146 UTC [1] LOG:  listening on IPv6 address "::", port 5432
db_1   | 2020-05-26 22:53:06.153 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db_1   | 2020-05-26 22:53:06.168 UTC [44] LOG:  database system was shut down at 2020-05-26 22:53:06 UTC
db_1   | 2020-05-26 22:53:06.172 UTC [1] LOG:  database system is ready to accept connections
app_1  | ! Unable to load application: KeyError: key not found: "SMTP_SENDER_EMAIL"
app_1  | bundler: failed to load command: puma (/usr/local/bundle/bin/puma)
app_1  | KeyError: key not found: "SMTP_SENDER_EMAIL"
app_1  |   /opt/rails/config/initializers/devise.rb:21:in `fetch'
app_1  |   /opt/rails/config/initializers/devise.rb:21:in `block in <main>'
app_1  |   /usr/local/bundle/gems/devise-4.7.1/lib/devise.rb:311:in `setup'
app_1  |   /opt/rails/config/initializers/devise.rb:5:in `<main>'
app_1  |   /usr/local/bundle/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:55:in `load'
app_1  |   /usr/local/bundle/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:55:in `load'
app_1  |   /usr/local/bundle/gems/activesupport-6.0.3.1/lib/active_support/dependencies.rb:318:in `block in load'
app_1  |   /usr/local/bundle/gems/activesupport-6.0.3.1/lib/active_support/dependencies.rb:291:in `load_dependency'
app_1  |   /usr/local/bundle/gems/activesupport-6.0.3.1/lib/active_support/dependencies.rb:318:in `load'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:666:in `block in load_config_initializer'
app_1  |   /usr/local/bundle/gems/activesupport-6.0.3.1/lib/active_support/notifications.rb:182:in `instrument'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:665:in `load_config_initializer'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:625:in `block (2 levels) in <class:Engine>'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:624:in `each'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:624:in `block in <class:Engine>'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:32:in `instance_exec'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:32:in `run'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:61:in `block in run_initializers'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:228:in `block in tsort_each'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:350:in `block (2 levels) in each_strongly_connected_component'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:422:in `block (2 levels) in each_strongly_connected_component_from'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:431:in `each_strongly_connected_component_from'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:421:in `block in each_strongly_connected_component_from'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:50:in `each'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:50:in `tsort_each_child'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:415:in `call'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:415:in `each_strongly_connected_component_from'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:349:in `block in each_strongly_connected_component'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:347:in `each'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:347:in `call'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:347:in `each_strongly_connected_component'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:226:in `tsort_each'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:205:in `tsort_each'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:60:in `run_initializers'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/application.rb:363:in `initialize!'
app_1  |   /opt/rails/config/environment.rb:5:in `<top (required)>'
app_1  |   config.ru:3:in `require_relative'
app_1  |   config.ru:3:in `block in <main>'
app_1  |   /usr/local/bundle/gems/rack-2.2.2/lib/rack/builder.rb:116:in `eval'
app_1  |   /usr/local/bundle/gems/rack-2.2.2/lib/rack/builder.rb:116:in `new_from_string'
app_1  |   /usr/local/bundle/gems/rack-2.2.2/lib/rack/builder.rb:105:in `load_file'
app_1  |   /usr/local/bundle/gems/rack-2.2.2/lib/rack/builder.rb:66:in `parse_file'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/configuration.rb:321:in `load_rackup'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/configuration.rb:246:in `app'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/runner.rb:155:in `load_and_bind'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/single.rb:98:in `run'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/launcher.rb:172:in `run'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/cli.rb:80:in `run'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/bin/puma:10:in `<top (required)>'
app_1  |   /usr/local/bundle/bin/puma:23:in `load'
app_1  |   /usr/local/bundle/bin/puma:23:in `<top (required)>'
simple_group_app_1 exited with code 1
```

Then I have to CTRL+C to exit out of that. 

Subsequent runs of `docker-compose -f docker-compose.development.yml --env-file .env up --build` end in this:

```
Starting simple_group_db_1 ... done
Starting simple_group_app_1 ... done
Attaching to simple_group_db_1, simple_group_app_1
db_1   | 
db_1   | PostgreSQL Database directory appears to contain a database; Skipping initialization
db_1   | 
db_1   | 2020-05-26 22:56:20.420 UTC [1] LOG:  starting PostgreSQL 12.3 on x86_64-pc-linux-musl, compiled by gcc (Alpine 9.2.0) 9.2.0, 64-bit
db_1   | 2020-05-26 22:56:20.421 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
db_1   | 2020-05-26 22:56:20.421 UTC [1] LOG:  listening on IPv6 address "::", port 5432
db_1   | 2020-05-26 22:56:20.430 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db_1   | 2020-05-26 22:56:20.472 UTC [20] LOG:  database system was shut down at 2020-05-26 22:55:55 UTC
db_1   | 2020-05-26 22:56:20.477 UTC [1] LOG:  database system is ready to accept connections
app_1  | Puma starting in single mode...
app_1  | * Version 4.3.5 (ruby 2.7.1-p83), codename: Mysterious Traveller
app_1  | * Min threads: 5, max threads: 5
app_1  | * Environment: development
app_1  | ! Unable to load application: KeyError: key not found: "SMTP_SENDER_EMAIL"
app_1  | bundler: failed to load command: puma (/usr/local/bundle/bin/puma)
app_1  | KeyError: key not found: "SMTP_SENDER_EMAIL"
app_1  |   /opt/rails/config/initializers/devise.rb:21:in `fetch'
app_1  |   /opt/rails/config/initializers/devise.rb:21:in `block in <main>'
app_1  |   /usr/local/bundle/gems/devise-4.7.1/lib/devise.rb:311:in `setup'
app_1  |   /opt/rails/config/initializers/devise.rb:5:in `<main>'
app_1  |   /usr/local/bundle/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:55:in `load'
app_1  |   /usr/local/bundle/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:55:in `load'
app_1  |   /usr/local/bundle/gems/activesupport-6.0.3.1/lib/active_support/dependencies.rb:318:in `block in load'
app_1  |   /usr/local/bundle/gems/activesupport-6.0.3.1/lib/active_support/dependencies.rb:291:in `load_dependency'
app_1  |   /usr/local/bundle/gems/activesupport-6.0.3.1/lib/active_support/dependencies.rb:318:in `load'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:666:in `block in load_config_initializer'
app_1  |   /usr/local/bundle/gems/activesupport-6.0.3.1/lib/active_support/notifications.rb:182:in `instrument'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:665:in `load_config_initializer'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:625:in `block (2 levels) in <class:Engine>'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:624:in `each'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/engine.rb:624:in `block in <class:Engine>'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:32:in `instance_exec'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:32:in `run'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:61:in `block in run_initializers'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:228:in `block in tsort_each'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:350:in `block (2 levels) in each_strongly_connected_component'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:422:in `block (2 levels) in each_strongly_connected_component_from'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:431:in `each_strongly_connected_component_from'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:421:in `block in each_strongly_connected_component_from'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:50:in `each'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:50:in `tsort_each_child'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:415:in `call'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:415:in `each_strongly_connected_component_from'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:349:in `block in each_strongly_connected_component'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:347:in `each'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:347:in `call'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:347:in `each_strongly_connected_component'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:226:in `tsort_each'
app_1  |   /usr/local/lib/ruby/2.7.0/tsort.rb:205:in `tsort_each'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/initializable.rb:60:in `run_initializers'
app_1  |   /usr/local/bundle/gems/railties-6.0.3.1/lib/rails/application.rb:363:in `initialize!'
app_1  |   /opt/rails/config/environment.rb:5:in `<top (required)>'
app_1  |   config.ru:3:in `require_relative'
app_1  |   config.ru:3:in `block in <main>'
app_1  |   /usr/local/bundle/gems/rack-2.2.2/lib/rack/builder.rb:116:in `eval'
app_1  |   /usr/local/bundle/gems/rack-2.2.2/lib/rack/builder.rb:116:in `new_from_string'
app_1  |   /usr/local/bundle/gems/rack-2.2.2/lib/rack/builder.rb:105:in `load_file'
app_1  |   /usr/local/bundle/gems/rack-2.2.2/lib/rack/builder.rb:66:in `parse_file'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/configuration.rb:321:in `load_rackup'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/configuration.rb:246:in `app'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/runner.rb:155:in `load_and_bind'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/single.rb:98:in `run'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/launcher.rb:172:in `run'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/cli.rb:80:in `run'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/bin/puma:10:in `<top (required)>'
app_1  |   /usr/local/bundle/bin/puma:23:in `load'
app_1  |   /usr/local/bundle/bin/puma:23:in `<top (required)>'
simple_group_app_1 exited with code 1
```

ashpil: I'm running this on Fedora 32, if that helps. I had to do some fiddling because docker doesn't technically have an official repo for Fedora 32 yet, but there's a workaround online that lets you run docker, and it should work perfectly.

ashpil: Also, this is odd. 

```
[ashpil@localhost simple_group]$ docker image ls
REPOSITORY                      TAG                 IMAGE ID            CREATED             SIZE
hernanat/simple-group-app-dev   latest              5d7917b3f5c6        9 minutes ago       266MB
<none>                          <none>              db183ae8f8f2        9 minutes ago       305MB
postgres                        12.3-alpine         bd3eeea38a5a        10 days ago         151MB
ruby                            2.7.1-alpine3.11    564510d09024        4 weeks ago         51.5MB
```

ashpil: I guess this backs up your argument for separate containers. Postgres seems to work just fine, it's the rails app (I assume) that is being wonky.

hernanat: do you have a `.env` file locally? use `.env.sample` as a template

hernanat:
as it turns out EB actually uses nginx as a reverse proxy by default and I can configure that itself. I'm going to look into doing that which will simplify this a great deal.

hernanat:
yet another option is using CloudFront as a CDN to serve static assets. The free-tier usage limits are 2million requests / month and after that it's like a penny per request...but then we're opening ourselves up to having a potentially costly monthly bill (if we surpass 2 million requests in a month and then get 2million more that's a heft bill).

I do think that the best path forward in the meantime is to actually configure the nginx reverse proxy that comes by default and reduce the complexity with all of this. i.e. we would just be using a single docker environment instead of a multi docker environment, still with nginx sitting in front of our app.

hernanat:
maybe we can collect some (anonymous -- literally just request counts) usage statistics and in the future if we want to move onto a CDN we can

ashpil:
> I do think that the best path forward in the meantime is to actually configure the nginx reverse proxy that comes by default and reduce the complexity with all of this. i.e. we would just be using a single docker environment instead of a multi docker environment, still with nginx sitting in front of our app.

That sounds good, simplicity is good. Wouldn't it still be a multi-container docker environment with the separation of postgres and rails, though?

docker-compose.development.yml
ashpil: Interesting...

By what you said, I assume I could just make a copy of `.env.sample` called `.env`, so I did that.

That gets me an error that looks essentially the same, but is much briefer:

```
Successfully built e8399239cf42
Successfully tagged hernanat/simple-group-app-dev:latest
Creating simple_group_db_1 ... done
Creating simple_group_app_1 ... done
Attaching to simple_group_db_1, simple_group_app_1
db_1   | 
db_1   | PostgreSQL Database directory appears to contain a database; Skipping initialization
db_1   | 
db_1   | 2020-05-28 03:55:10.205 UTC [1] LOG:  starting PostgreSQL 12.3 on x86_64-pc-linux-musl, compiled by gcc (Alpine 9.2.0) 9.2.0, 64-bit
db_1   | 2020-05-28 03:55:10.206 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
db_1   | 2020-05-28 03:55:10.206 UTC [1] LOG:  listening on IPv6 address "::", port 5432
db_1   | 2020-05-28 03:55:10.216 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db_1   | 2020-05-28 03:55:10.239 UTC [20] LOG:  database system was shut down at 2020-05-26 22:59:53 UTC
db_1   | 2020-05-28 03:55:10.244 UTC [1] LOG:  database system is ready to accept connections
app_1  | Puma starting in single mode...
app_1  | * Version 4.3.5 (ruby 2.7.1-p83), codename: Mysterious Traveller
app_1  | * Min threads: 5, max threads: 5
app_1  | * Environment: development
app_1  | * Listening on tcp://0.0.0.0:3000
app_1  | bundler: failed to load command: puma (/usr/local/bundle/bin/puma)
app_1  | Errno::ENOENT: No such file or directory @ rb_sysopen - tmp/pids/server.pid
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/launcher.rb:216:in `initialize'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/launcher.rb:216:in `open'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/launcher.rb:216:in `write_pid'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/launcher.rb:105:in `write_state'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/single.rb:103:in `run'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/launcher.rb:172:in `run'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/lib/puma/cli.rb:80:in `run'
app_1  |   /usr/local/bundle/gems/puma-4.3.5/bin/puma:10:in `<top (required)>'
app_1  |   /usr/local/bundle/bin/puma:23:in `load'
app_1  |   /usr/local/bundle/bin/puma:23:in `<top (required)>'
simple_group_app_1 exited with code 1
```

hernanat:
> That sounds good, simplicity is good. Wouldn't it still be a multi-container docker environment with the separation of postgres and rails, though?

hernanat:
> That sounds good, simplicity is good. Wouldn't it still be a multi-container docker environment with the separation of postgres and rails, though?

I'm not sure if they have nginx also running on a docker container within the machine, or running directly on the machine. Actually I assume on the machine. We can configure that instance of nginx itself so I think it's best to just do that instead of introducing yet another.

As for the database, it's running on its own machine entirely. I know I mentioned the three container scenario when I was talking about Docker but I was more just trying to illustrate the value of keeping one process per container. In production you typically don't want to have your database running on the same machine as your application for a variety of reasons. Also AWS RDS takes care of the desired redundancy and backups pertaining to databases so we don't have to (we would have to figure that out if we were running the database ourselves).

hernanat:
I really meant to work on this today but I didn't get a chance. Was dealing with some things and then I also wasn't feeling very well. I'm going to try to finish this up tomorrow and I might actually open up a separate PR (or PRs to keep this more atomic and to help me think more specifically)

docker-compose.development.yml
ashpil: Any thoughts on this, @hernanat?

hernanat: hm. this is actually an error I saw previously. let me try to remember how I fixed it. I think it might have something to do with the `tmp` directory being ignored and so from the container's perspective it doesn't exist when it tries to create the server pid in it

hernanat: can you try running: `bundle exec puma -C config/puma.rb`

first?

hernanat:
I'm _extremely_ close to giving up on elastic beanstalk. This is fucking ridiculous. It's like they looked at everything that other hosting platforms were doing to streamline deployments and went: "you know what, we're going to do the _exact opposite_."

It _should not_ be this difficult to come up with a reproducible / automatable deployment setup. Yet Amazon always finds a way. I'm not even going to get start on the complete bastardization of `docker-compose` that the `eb` config is sitting on top of. Fucking dumpster fire.

docker-compose.development.yml
ashpil: From the container, presumably? Should I just put it before the final command in the Dockerfile?

hernanat: No locally first

hernanat: To see if it runs

ashpil: Outside the container gets me this:

```
Ignoring fast_blank-1.0.0 because its extensions are not built. Try: gem pristine fast_blank --version 1.0.0
Ignoring ffi-1.12.2 because its extensions are not built. Try: gem pristine ffi --version 1.12.2
Ignoring sassc-2.2.1 because its extensions are not built. Try: gem pristine sassc --version 2.2.1
bundler: command not found: puma
Install missing gem executables with `bundle install`
```
Keep in mind I typically run this on my Windows machine, but I've been testing the docker version on my Linux one. This machine knows nothing about the existence of ruby/etc. I assume this shouldn't be an issue with docker.

hernanat: i.e. you're running the command in the docker container? Wherever you run it should have the full app environment configured. The command starts the web server. My suspicion is that puma is trying to add its server pid within `tmp/pids` but the `pids` folder doesn't exist

hernanat:
@ashpil I'm going to close this and open up a separate PR for the dev environment docker stuff. When I do that, you and I should setup some time to try and get it up and running for you.

PR #50, opened 2020-06-20
require invitation + allow users to invite

hernanat:
this commit changes the registration workflow to require that users
  are invited in order to register. it also allows current users to
  send out invitations.


hernanat:
the workflow is similar to lobsters in that the user sends the invitations from their profile edit page wasn't sure of a decent place to put this otherwise.

PR #55, opened 2020-08-22
flash polish

ashpil:
This is a small PR that changes the flash frontend a bit.

- Flashes are less buggy in CSS:
    - Multiple flashes stack properly
    - "x" stays in right(or left, really) place on smaller screens
    - Flashes properly float on top of latex
- Now using JS instead of  the hacky input/checkbox workaround for closing the flashes
    - Simpler HTML/CSS
    - A bit more complicated javascipt (but now timeouts actually exist!)

### DIFF
--- a/app/assets/stylesheets/base/_flash.scss
+++ b/app/assets/stylesheets/base/_flash.scss
@@ -1,22 +1,31 @@
-.flash-message {
-    cursor: pointer;
+#flash-container {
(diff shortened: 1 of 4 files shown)


hernanat:
can you rebase / squash your changes here and on your other branch? Once these are both in I think I want to do a deploy today

PR #56, opened 2020-08-22
polishing of `new submission` page

ashpil:
Makes the new submission page have a tabbed structure for body/URL, which makes it clearer that you only need one. I'd attach a gif but I haven't figured out how to do that on Linux + Wayland yet. I'm not 100% satisfied with the CSS yet.

My wishlist for this:
 - Remove all `*` stars that indicate field requirement, since they aren't really necessary for clarity anymore with this tabbed structure(all visible fields are required), and are just clutter.
 - Save state on error - right now currently selected tab, as well as information in other tab, is lost on error because all of that is client-side. This can be somewhat mitigated by introducing client-side HTML5 form validations as well, but it wouldn't solve the issue.

I am not certain how to accomplish the above two tasks, hence this WIP PR.

### DIFF
--- a/app/assets/stylesheets/base/_variables.scss
+++ b/app/assets/stylesheets/base/_variables.scss
@@ -1,6 +1,7 @@
 // Colors
 $alto: #cecece;
 $blue: #1565c0;
+$blue-medium: #0e7ab6;
 $blue-light: #3b83d5;
 $blue-light-very: #f4fbff;
 $congress-blue: #004e98;
--- a/app/views/submissions/new.html.haml
+++ b/app/views/submissions/new.html.haml
@@ -1,14 +1,26 @@
+- content_for :page_styles do
+  = javascript_pack_tag "submission_new"
 %h2.prompt= t(".title")
-%ul.errors
-  - @form.errors.values.flatten.map do |e|
-    %li.error= e
 = simple_form_for @form, url: submissions_path do |f|
-  = f.input :url
-  = f.input :title
+  = f.input :title, input_html: { minlength: 10, maxlength: 175 }
   = f.input :tag_ids, collection: tag_select_options(current_user),
     input_html: { multiple: true, id: "tag-select" }
-  = f.input :body, as: :text, input_html: { rows: 15 }
+  %label= t(".submission_type")
+  #type-selector
+    %input{ type: "radio", id: "link", name: "option", checked: "checked" }
+    %label{ for: "link" }= t(".link_type")
+    %input{ type: "radio", id: "text", name: "option" }
+    %label{ for: "text" }= t(".text_type")
+  = f.input :url, wrapper_html: { id: "url_wrapper" }
+  = f.input :body, as: :text, input_html: { rows: 15, disabled: "disabled" },
+    wrapper_html: { style: "display:none;", id: "body_wrapper" }
   .original-author-statement
     = f.input :original_author, as: :boolean, wrapper: :left_hand_label
     = t(".original_author")
   = f.button :button, type: :submit
+  - if @form.errors.messages.key?(:user)
+    %span.error= @form.errors.messages[:user].join("")
+  - if @form.errors.messages.key?(:base)
+    %span.error= @form.errors.messages[:base].join("")
+  - if @form.errors.messages.key?(:domain)
+    %span.error= @form.errors.messages[:domain].join("")
--- a/spec/support/page_objects/create_submission_page.rb
+++ b/spec/support/page_objects/create_submission_page.rb
@@ -22,6 +22,20 @@ class CreateSubmissionPage < PageBase
     self
   end
 
+  def click_link_tab
+    within("div#type-selector") do
+      find("label", text: t("submissions.new.link_type")).click
+    end
+    self
+  end
+
+  def click_text_tab
+    within("div#type-selector") do
+      find("label", text: t("submissions.new.text_type")).click
+    end
+    self
+  end
+
   def fill_in_body(str)
     fill_in :create_submission_form_body, with: str
     self
@@ -39,12 +53,27 @@ class CreateSubmissionPage < PageBase
     self
   end
 
-  def has_title_missing_error?
-    has_error?(t("submissions.new.error.title_missing"))
+  def has_submission_body_text_box?
+    has_css?("#body_wrapper", visible: true)
   end
 
-  def has_title_length_error?
-    has_error?(t("submissions.new.error.title_length"))
+  def has_submission_url_text_box?
+    has_css?("#url_wrapper", visible: true)
+  end
+
+  def has_title_missing_error?
+    has_client_side_error?("#create_submission_form_title", "Please fill out this field.")
+  end
+
+  def has_title_too_short_error?(char_num)
+    has_client_side_error?(
+      "#create_submission_form_title",
+      "Please lengthen this text to 10 characters or more (you are currently using #{char_num} characters)."
+    )
+  end
+
+  def has_title_with_x_chars?(char_num)
+    page.find("#create_submission_form_title").value.length == char_num
   end
 
   def has_url_xor_body_error?
@@ -52,7 +81,7 @@ class CreateSubmissionPage < PageBase
   end
 
   def has_missing_tags_error?
-    has_error?(t("submissions.new.error.tags_missing"))
+    has_client_side_error?("#tag-select", "Please select an item in the list.")
   end
 
   def has_tags_max_error?
@@ -88,8 +117,12 @@ class CreateSubmissionPage < PageBase
   end
 
   def has_error?(error)
-    within find(".errors") do
-      has_css?("li.error", text: error)
+    within("form#new_create_submission_form") do
+      has_css?("span.error", text: error)
     end
+  end
+
+  def has_client_side_error?(field, message)
+    page.find(field).native.attribute("validationMessage") == message
   end
 end
--- a/spec/system/user_makes_submission_spec.rb
+++ b/spec/system/user_makes_submission_spec.rb
@@ -18,6 +18,22 @@ RSpec.describe "user makes a submission" do
   context "when the user is signed in" do
     let(:user) { create(:user) }
 
+    context "when the user switches between URL and text tabs", js: true do
+      it "displays the correct form elements" do
+        page.visit(as: user).
+          click_text_tab
+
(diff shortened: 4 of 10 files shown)


hernanat:
> Not sure about these last two leftover failures.
> 
>     1. Why does this go through?? Normally, the client-side validation for making sure an input doesn't go over a certain amount is just not let anything to be typed past a certain threshold. Evidently, the tests are able to bypass this. But, then why doesn't the server catch this??

why is what going through exactly?

>     2. Not sure if this test is necessary anymore. Body and URL are never able to be submitted together, as it just submits whatever type is active. I guess it would be nice to test the post request/backend and make sure the server rejects it, but the client wouldn't experience this issue so I'm unsure how to test that.

we have a test `spec/form_objects/create_submission_form_spec.rb` that handles the scenario of both a URL and body being present in the submission form object. We could also add a request spec if we wanted to.

spec/system/user_makes_submission_spec.rb
```diff
@@ -18,6 +18,18 @@
   context "when the user is signed in" do
     let(:user) { create(:user) }
 
+    context "when the user switches between URL and text tabs" do
+      it "displays the correct form elements" do
```
hernanat, line 22: you forgot to visit the page, so this spec will fail.

```ruby
page.visit(as: user)
```

spec/system/user_makes_submission_spec.rb
```diff
@@ -18,6 +18,18 @@
...
+        page.click_text_tab
```
hernanat, line 23: style nit: when we write specs we typically separate by new line the following sections:

- setup
- actions
- expecations

so it would be:

- setup
- visit + click tab
- expect

and then separate the next section with a new line

spec/system/user_makes_submission_spec.rb
ashpil: Huh, it does seem to succeed, though.

spec/system/user_makes_submission_spec.rb
```diff
@@ -18,6 +18,18 @@
...
+        expect(page).to have_submission_body_text_box
+        expect(page).not_to have_submission_url_text_box
+
+        page.click_url_tab
```
hernanat, line 27: `click_link_tab` ?

spec/system/user_makes_submission_spec.rb
```diff
@@ -18,6 +18,18 @@
   context "when the user is signed in" do
     let(:user) { create(:user) }
 
+    context "when the user switches between URL and text tabs" do
```
hernanat, line 21: need `js: true` here

ashpil:
> why is what going through exactly?

I'm referring to the two leftover failures in CircleCI. The first failure is a submission going through with 176 characters.

spec/system/user_makes_submission_spec.rb
ashpil: Should I separate the body/url text box tests into separate tests? Or is it fine to go over those two sections twice?

spec/system/user_makes_submission_spec.rb
```diff
@@ -221,12 +237,14 @@
       end
     end
 
-    context "when both the URL and body are present" do
-      it "displays an error at the top of the page and doesn't save the submission" do
+    context "when both the URL and body are present", js: true do
```
hernanat, line 240: we can get rid of this test case

app/views/submissions/new.html.haml
```diff
@@ -1,11 +1,8 @@
 - content_for :page_styles do
   = javascript_pack_tag "submission_new"
 %h2.prompt= t(".title")
-%ul.errors
-  - @form.errors.values.flatten.map do |e|
-    %li.error= e
 = simple_form_for @form, url: submissions_path do |f|
-  = f.input :title
+  = f.input :title, input_html: { minlength: 10, maxlength: 175 }
```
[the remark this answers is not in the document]
hernanat: yeah so it appears to be the case that only the `minlength` property will generate a client side error when you submit the form. The behavior of `maxlength` appears to be that it will just prevent the user from entering any more characters past 175.

spec/support/page_objects/create_submission_page.rb
```diff
@@ -50,19 +54,19 @@ def submit_form
   end
 
   def has_title_missing_error?
-    has_error?(t("submissions.new.error.title_missing"))
+    has_client_side_error?("#create_submission_form_title", "Please fill out this field.") || has_error?(t("submissions.new.error.title_missing"))
```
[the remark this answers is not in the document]
hernanat: we really ought to have just one or the other. I can't really see a situation where the clientside validation would fail, probably an edgecase that we don't need to worry much about.

spec/support/page_objects/create_submission_page.rb
```diff
@@ -50,19 +54,19 @@ def submit_form
...
   end
 
-  def has_title_length_error?
-    has_error?(t("submissions.new.error.title_length"))
+  def has_title_length_error?(char_num)
+    has_client_side_error?("#create_submission_form_title", "Please lengthen this text to 10 characters or more (you are currently using #{char_num} characters).") || has_error?(t("submissions.new.error.title_length"))
```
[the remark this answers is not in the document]
hernanat: we would have to localize it on the client side, which at this point I don't really feel like doing :innocent:

app/assets/stylesheets/base/_variables.scss
```diff
@@ -1,6 +1,7 @@
 // Colors
 $alto: #cecece;
 $blue: #1565c0;
+$blue-medium: #0e7ab6;
```
hernanat, line 4: idk why it isn't showing up in the diff here, but there is a trailing whitespace on this like (if you run `git diff` or `git diff origin/master` you can see it

spec/system/user_makes_submission_spec.rb
ashpil: Yeah, that's what I was thinking.

app/views/submissions/new.html.haml
```diff
@@ -19,4 +19,8 @@
     = t(".original_author")
   = f.button :button, type: :submit
   - if @form.errors.messages.key?(:user)
-    .error= @form.errors.messages[:user].join("")
+    %span.error= @form.errors.messages[:user].join("")
+  - if @form.errors.messages.key?(:base)
+    %span.error= @form.errors.messages[:base].join("")
+  - if @form.errors.messages.key?(:domain)
+    %span.error= @form.errors.messages[:domain].join("")
```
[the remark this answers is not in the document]
hernanat: `domain` and `url` are not the same thing -- the `domain` is the literal domain that the URL is hosted on. We could have many URLs for a single domain. The usecase here is that if we determine that a domain is malicious / giving out false information / whatever reason, we can blacklist all URLs from it.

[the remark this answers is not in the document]
hernanat: we could also just do something like:

```ruby
@form.errors.full_messages.each do |err|
  # do something with the error text
end
```

I'm indifferent

spec/system/user_makes_submission_spec.rb
hernanat: you can just keep them in here

hernanat:
image

hernanat:
errr woops didn't mean to screenshot both my screens

hernanat:
edit: wrong PR. wow I'm dumb

app/views/submissions/new.html.haml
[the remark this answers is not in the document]
ashpil: Then why is the test still able to go past 175? Shouldn't that give a server-side validation error?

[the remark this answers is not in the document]
hernanat: is it persisting a title with more than 175 characters? The test simulates browser interactions, so it's probably limiting the input to 175 characterrs, and so in the test if you're attempting to input 176 `f`s, what will be persisted in the database is 175 `f`s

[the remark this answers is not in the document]
ashpil: Oh nvm I get it, the machine is only able to submit 175 characters.

spec/system/user_makes_submission_spec.rb
```diff
@@ -182,24 +196,26 @@
...
 
         expect(Submission.count).to eq(0)
-        expect(page).to have_title_length_error
+        expect(page).to have_title_length_error("8")
       end
     end
 
-    context "when the title is too long" do
-      it "displays an error at the top of the page and doesn't save the submission" do
+    context "when the title is too long", js: true do
```
hernanat, line 213: this should either get refactored or removed given what we know now. We won't ever get a "too long" error by submitting the form, so maybe we should change this to test that you can't input past 175 characters?

spec/system/user_makes_submission_spec.rb
```diff
@@ -206,22 +211,16 @@
...
-          select_tag(tag.id).
-          click_text_tab.
-          fill_in_body("This is a very nice submission body with some great info").
-          check_original_author.
-          submit_form
+          fill_in_title("f" * 176)
 
-        expect(Submission.count).to eq(0)
-        expect(page).to have_title_length_error
+        expect(page).to have_title_with_x_chars(175)
```
ashpil, line 223: Does this simplified version make sense? I wasn't sure what the proper way to do this was...

hernanat: yeah I think this is fine

hernanat:
@ashpil I would do your other PR first and then switch the base on this one back to master, squash + rebase, then we can merge

ashpil:
> 
> 
> @ashpil I would do your other PR first and then switch the base on this one back to master, squash + rebase, then we can merge

Other is already done, I think

ashpil:
Whoops I guess I was supposed to squash the flash-rework one too. Lemme do that.

hernanat:
@ashpil remove all the flash stuff from this branch and then rebase on top of master and it should be fine

ashpil:
> @ashpil remove all the flash stuff from this branch and then rebase on top of master and it should be fine

Honestly, after an attempt, not sure how to do that properly. Would you mind doing it? You could probably do it much faster anyway :p

hernanat:
sure

hernanat:
@ashpil does this look correct to you?

ashpil:
Think so

hernanat:
:ship: :rocket:

PR #57, opened 2020-08-23
allows users to cancel their accounts

hernanat:
https://trello.com/c/Yhd01aZw

This commit adds functionality for users to cancel their accounts
  and *mostly* remove their data. The strategy is as follows:

- all references to username / email are set to `[deleted]`, including:
  - account email
  - unconfirmed emails
  - the email on their original invitation to the site if it exists

- all submission text / comment bodies are set to `[deleted]`

- account password is set to a secure random uuid

- account role is set to `deactivated`

The intent is to maintain the integrity of our data (so that we don't do
  things like mess up the invite tree, comment trees, screw up user
  karma, etc), while also giving users the ability to more or less scrub
  themselves from our site.

We don't keep a lot of personally identifying information at present, so
  things are pretty straightforward thus far. In the future, when we
  allow users to do things like share their contact information (i.e.
  social media profiles, githubs, etc.) we can just change the removal
  service to scrub those as well.

I did not do anything relating to submission titles / URL submissions,
  the rationale being:

- titles (in the future) can be auto-generated given a link
  - also, a submission title being used as a personally identifiable
    piece of information is questionable at best, in general

- URLs are on the internet regardless, and so public

On top of the above, we introduce the following behaviors:

- signups with the username `[deleted]` are not allowed

- attempting to signin as `[deleted]` will not even entertain an auth
  request, it will just redirect to the signin page

- attempting to signin as a deactivated account will display an invalid
  credentials error

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -1,4 +1,11 @@
 # frozen_string_literal: true
 
 class Users::SessionsController < Devise::SessionsController
+  def create
+    if sign_in_params[:username].downcase == User::DELETED
+      redirect_to new_user_session_path
+    else
+      super
+    end
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
+  DELETED = "[deleted]"
+
   INVITES_PER_DAY = 10
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
@@ -11,7 +13,8 @@ class User < ApplicationRecord
   enum role: {
     member: 0,
     moderator: 10,
-    admin: 20
+    admin: 20,
+    deactivated: 100
   }
 
   has_many :submissions
@@ -19,16 +22,19 @@ class User < ApplicationRecord
   has_many :comments
   has_many :votes
   has_many :sent_user_invitations, foreign_key: :sender_id, class_name: :UserInvitation
+  has_one :received_user_invitation, foreign_key: :recipient_id, class_name: :UserInvitation, required: false
 
   validates :username, presence: true
-  validates :email, email: true
+  validate :username_not_restricted?
+
+  scope :active, -> { where.not(role: :deactivated) }
 
   def self.find_first_by_auth_conditions(warden_conditions)
     if warden_conditions.key?(:confirmation_token)
       super
     else
       warden_conditions.fetch(:username, nil).then do |username|
-        where("lower(username) = ?", username.downcase).first if username.present?
+        active.where("lower(username) = ?", username.downcase).first if username.present?
       end
     end
   end
@@ -57,5 +63,11 @@ class User < ApplicationRecord
     sent_user_invitations.
       where(sent_at: (now.beginning_of_day..now.end_of_day)).
       count < INVITES_PER_DAY
+  end
+
+  def username_not_restricted?
+    if username.present? && username.downcase == DELETED
+      errors.add(:username, :invalid)
+    end
   end
 end
--- a/app/services/user_removal_service.rb
+++ b/app/services/user_removal_service.rb
@@ -1,0 +1,55 @@
+class UserRemovalService
+  def initialize(user)
+    @user = user
+  end
+
+  def remove
+    ApplicationRecord.transaction do
+      remove_submission_data
+      remove_comment_data
+      remove_invitation_data
+      deactivate_user
+    end
+  end
+
+  private
+
+  # rubocop:disable Rails/SkipsModelValidations
+
+  def remove_submission_data
+    user.submissions.where(url: nil).update_all(body: User::DELETED)
+  end
+
+  def remove_comment_data
+    user.comments.update_all(body: User::DELETED)
+  end
+
+  def remove_invitation_data
+    user.received_user_invitation&.update_column(:recipient_email, User::DELETED)
+  end
+
+  def deactivate_user
+    password = SecureRandom.uuid
+
+    user.update!(
+      password: password,
+      password_confirmation: password
+    )
+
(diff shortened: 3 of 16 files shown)


hernanat:
[As a user I can delete my account](https://trello.com/c/Yhd01aZw/45-as-a-user-i-can-delete-my-account)

app/controllers/users/sessions_controller.rb
```diff
@@ -1,4 +1,11 @@
 # frozen_string_literal: true
 
 class Users::SessionsController < Devise::SessionsController
+  def create
+    if sign_in_params[:username].downcase == User::DELETED
```
hernanat, line 5: I've decided not to even entertain a database call / a call to the auth workflow if someone tries to login with this username.

app/models/user.rb
```diff
@@ -11,24 +13,28 @@ class User < ApplicationRecord
   enum role: {
     member: 0,
     moderator: 10,
-    admin: 20
+    admin: 20,
+    deactivated: 100
```
hernanat, line 17: I thought about adding a flag but then I figured if a user is deactivated it doesn't really matter what their role was before they were deactivated. If we ever reactivate them then we can set their role to whatever.

app/models/user.rb
```diff
@@ -11,24 +13,28 @@ class User < ApplicationRecord
...
   }
 
   has_many :submissions
   has_many :submission_actions
   has_many :comments
   has_many :votes
   has_many :sent_user_invitations, foreign_key: :sender_id, class_name: :UserInvitation
+  has_one :received_user_invitation, foreign_key: :recipient_id, class_name: :UserInvitation, required: false
```
hernanat, line 25: a user will only ever have one received invitation, but certain users (root users) won't have one at all.

app/models/user.rb
```diff
@@ -11,24 +13,28 @@ class User < ApplicationRecord
...
 
   validates :username, presence: true
-  validates :email, email: true
+  validate :username_not_restricted?
+
+  scope :active, -> { where.not(role: :deactivated) }
```
hernanat, line 30: we use this below to exclude deactivated accounts from our auth workflow

app/services/user_removal_service.rb
```diff
@@ -0,0 +1,55 @@
...
+      remove_submission_data
+      remove_comment_data
+      remove_invitation_data
+      deactivate_user
+    end
+  end
+
+  private
+
+  # rubocop:disable Rails/SkipsModelValidations
```
hernanat, line 17: I'm skipping model validations here because:

1.) I want to update in bulk anyways, and model validations don't apply to bulk actions anyways

2.) we have model validations that do things like validate the email format, so `[deleted]` would fail. But there is no need to have a "valid" email.

app/services/user_removal_service.rb
```diff
@@ -0,0 +1,55 @@
...
+  def remove_submission_data
+    user.submissions.where(url: nil).update_all(body: User::DELETED)
+  end
+
+  def remove_comment_data
+    user.comments.update_all(body: User::DELETED)
+  end
+
+  def remove_invitation_data
+    user.received_user_invitation&.update_column(:recipient_email, User::DELETED)
```
hernanat, line 28: `&` safe navigation because the user might not have an invitation at all

app/services/user_removal_service.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def deactivate_user
+    password = SecureRandom.uuid
+
+    user.update!(
```
hernanat, line 34: I'm updating their password *just in case*, whether it be they find some weird way to log back in after, or if there is ever some sort of data leak (god I hope not), at least if there are any hashed passwords leaked, users who have cancelled their accounts don't have to worry about it being a hash of their own password.

db/migrate/20200823183710_change_uniqueness_indexes_on_users_to_exclude_deleted.rb
```diff
@@ -0,0 +1,19 @@
+class ChangeUniquenessIndexesOnUsersToExcludeDeleted < ActiveRecord::Migration[6.0]
+  def up
+    remove_index :users, name: "index_users_on_LOWER_email"
+    remove_index :users, name: "index_users_on_LOWER_username"
+
+    add_index :users, "LOWER(email)", unique: true, name: "index_users_on_LOWER_email",
```
hernanat, line 6: here and elsewhere, what we're doing is we're excluding `[deleted]` from our uniqueness constraint

spec/system/user_cancels_account_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    page = UserEditPage.new
+
+    page.visit(as: user)
+
+    page.cancel_account
+
+    expect(page).to have_current_path(root_path)
+    expect(page).to have_account_cancelled_notice
+
+    user.reload
```
hernanat, line 19: this line and down is basically verbatim the expectations from the service spec itself. Originally I had tried to just stub out the service and verify that it had been called, but for whatever reason couldn't get that to work. In any case, this doesn't really hurt at all, and it keeps things explicit.

app/models/user.rb
ashpil: That's us!

hernanat: we're the cool kidz :sunglasses:

hernanat:
> I assume, currently, if someone deletes their account, their username is up for grabs again, right? So if `NumberphileOfficial` deleted their account, someone else could come grab it? I assume that wouldn't be an issue since they wouldn't inherit other info, but it does feel sort of weird. I have no idea how other platforms handle this, if at all.

Yeah as of now that would be the case. In the future we could introduce some sort of verification process if we need to for certain users but as of this moment I feel like that'd be some sort of weird edge case. I would love to have them sign up and NOT delete their account though :P

hernanat:
thanks @ashpil

PR #59, opened 2020-08-23
[WIP] notifications + inbox items

hernanat:


hernanat:
@ashpil id if you have style preferences

ashpil:
Put some pics up of the WIP?

hernanat:
image

hernanat:
image

@ashpil this is what I currently have. tldr; an "All" feed for all types of inbox items, and then two tabs (so far) for comment / submission replies (similar to reddit). Nothing fancy.

hernanat:
I haven't implemented the views for the two tabs yet. Idk what we really want this design to look like. The "All" feed will presumably have the things it currently has, and then things like name mentions, removal notifications, whatever in the future. In order to make this not be terribly inefficient I've had to introduce another database view, which we would expand on in the future too I guess. I found it incredibly difficult to come up with good / general naming conventions for the different sorts of things that we need to include in it. I think it'll be easier once we hammer out a design and then the fields can correspond with what they're taken to be on the page.

ashpil:
Can we use the same comment view for the comment structure rather than creating another one?

Also, do we even have to think of submission/comment replies as different entities? Aren't they the same, really? Both replies? Not sure if we need a tabbed structure at all here.

hernanat:
> Can we use the same comment view for the comment structure rather than creating another one?

heh...yeah...that's what I want to do eventually but that would require some significant refactoring on my part (which I would like to do anyways). Recall that I did a really hacky thing to get comments displayed nicely on the page, so I would have to create a partial and then refactor the submissions view to use that partial, which from our previous experience does not seem as easy as it sounds.

> Also, do we even have to think of submission/comment replies as different entities? Aren't they the same, really? Both replies? Not sure if we need a tabbed structure at all here.

I was going off what Reddit has to try to make things seem more familiar. I do think there is value to being able to say "show me all the replies to my *comments*" vs "show me all the direct replies to my *submissions*", at least I would find something like that helpful.

ashpil:
> heh...yeah...that's what I want to do eventually but that would require some significant refactoring on my part (which I would like to do anyways). Recall that I did a really hacky thing to get comments displayed nicely on the page, so I would have to create a partial and then refactor the submissions view to use that partial, which from our previous experience does not seem as easy as it sounds.

That would be very neat, as then we could have a preview of what comment of yours the reply is to.

> I was going off what Reddit has to try to make things seem more familiar. I do think there is value to being able to say "show me all the replies to my _comments_" vs "show me all the direct replies to my _submissions_", at least I would find something like that helpful.

If that's the use-case we're going for, it might make more sense to think of it as a filter, rather than a tabbed structure. Then we can also introduce more controls in the future, if need be.

> Show: 🗷 Replies to Comments 🗷 Replies to Posts

hernanat:
I don't mind a sprinkle of js here and there. If you think it would be a good fit here then we can roll with it. If you design something I'll implement it :P

ashpil:
Sweet, though I feel like it doesn't make a lot of sense to do a lot of designing here just yet, until we have the proper comment partial - otherwise we'd just end up replacing it.

I think just a type of 

> Show: 🗷 Replies to Comments 🗷 Replies to Posts

would work rather well, check what you want to show. Uncheck those you don't want to see. I guess the issue would be with ruby-js interop, if we do something like multiple pages of replies, how do we make sure only what they have selected shows on page update.

hernanat:
yeah the moreI think about it the more I think it might make more sense for me to tackle the partial issue first

hernanat:
> would work rather well, check what you want to show. Uncheck those you don't want to see. I guess the issue would be with ruby-js interop, if we do something like multiple pages of replies, how do we make sure only what they have selected shows on page update.

this is actually not too hard to do hehe

hernanat:
ok now to rework this for the new and improved comment partial

PR #60, opened 2020-08-30
refactors comment tree view code

hernanat:
- introduces comment partial

- render toplevel comments as collection then recursively
  render children

- eventually we might want to limit tree depth depending on how
  many comments / depth per comment we see per thread on average.


app/views/comments/_comment.html.haml
```diff
@@ -0,0 +1,25 @@
+- children = @comments_by_parent[comment.id]
```
hernanat, line 1: haml lint really doesn't like me doing this, but I disabled the cop for this file because I didn't feel like having to write `@comments_by_parent[comment.id]` repeatedly, and unfortunately there is not a way to define the children as a function of the parent when calling `render` (I looked and couldn't find anything at least, I might open up a PR in Rails for this)

app/views/comments/_comment_tree.html.erb
```diff
@@ -1,45 +0,0 @@
-<% subtree = comments_by_parent[nil] %>
```
hernanat, line 1: GOODBYE UGLY

app/views/submissions/show.html.haml
```diff
@@ -3,7 +3,7 @@
   = sanitize CommonMarker.render_html(@submission.body.gsub("\\", "\\\\\\")) if @submission.body.present?
 .comments-container
   = render "comments/form", comment: @root_comment, parent: nil
-  = render "comments/comment_tree", comments_by_parent: @comments_by_parent
+  = render partial: "comments/comment", collection: @comments_by_parent[nil]
```
hernanat, line 6: comments with `nil` as the parent id are toplevel comments

app/views/comments/_comment.html.haml
```diff
@@ -0,0 +1,25 @@
...
+    %div{ id: "downvote_#{comment.short_id}", class: control_classes(comment, :downvote) }
+  %span.hide-comment
+  .comment-line1
+    %span.author= link_to comment.commenter, "#"
+    %span= submitted_time_text(comment)
+    |
+    %span.link= link_to t(".link"), "##{comment.short_id}", data: { turbolinks: false }
+    |
+    %span.link= link_to t(".reply"), "", class: "reply"
+  - if children.present?
```
hernanat, line 16: in my notifications work I think I might add to this a local variable `render_children` which is a boolean that tells the partial whether or not to render the children. In the notifications view we wouldn't want to render the children, for example.

app/views/submissions/show.html.haml
hernanat, line 6: see [here](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-collections) for info on how this works

app/views/comments/_comment.html.haml
```diff
@@ -0,0 +1,25 @@
...
+    .comment-tree-line
+  .comment-content{ id: "comment_content_#{comment.short_id}" }
+    .comment-body.latex-container.markdownable{ id: "tree_#{comment.short_id}" }
+      = sanitize CommonMarker.render_html(comment.body.gsub("\\", "\\\\\\"))
+    .comment-reply-box{ id: "reply_#{comment.short_id}" }
+      = render "comments/form", comment: Comment.new, parent: comment
+    .comment-replies{ id: "replies_#{comment.short_id}" }
+      - if children.present?
+        = render partial: "comments/comment", collection: children
```
hernanat, line 25: eventually we might want to limit the depth of the children in some way, and then have a "view more" button to load more. Or maybe we'll never get to that point :man_shrugging: won't know until we launch and folks are using the site

hernanat:
thanks @ashpil

app/views/comments/_comment.html.haml
hernanat: I opened up a proposal for this in the Rails mailing list [here](https://discuss.rubyonrails.org/t/feature-proposal-allow-procs-lambdas-in-locals-for-render-calls-on-collections-to-support-dependent-variables/76141/2) unfortunately only one person has responded and they seem to have ignored part of what I was saying. Just tried to clarify...we'll see.

PR #61, opened 2020-08-31
[WIP] Polishes user edit page

ashpil:
Polishes the user edit page, mainly just rearranging a couple of things.

image

### DIFF
--- a/app/assets/stylesheets/base/_tables.scss
+++ b/app/assets/stylesheets/base/_tables.scss
@@ -1,6 +1,6 @@
 table {
   border-collapse: collapse;
-  margin: $spacing 0;
+  margin-bottom: $spacing;
   table-layout: fixed;
   text-align: left;
   width: 100%;
--- a/app/assets/stylesheets/pages/_users.scss
+++ b/app/assets/stylesheets/pages/_users.scss
@@ -1,1 +1,26 @@
 @import "users/edit";
+
+
+form {
+  .options {
+    span {
+      font-weight: 800;
+    }
+
+    span:last-child {
+      float: right;
+    }
+
+    .warning {
+      color: darken($red, 5%);
+    }
+  }
+
+  .hidden {
+    display: none;
+  }
+
+  .changing-password {
+    display: initial !important;
+  }
+}
--- a/app/views/users/registrations/edit.html.haml
+++ b/app/views/users/registrations/edit.html.haml
@@ -1,24 +1,38 @@
+- content_for :page_styles do
+  = javascript_pack_tag "user"
 %h2.prompt= t(".title", resource: resource.model_name.human)
+%h3= t(".edit_acount")
 = simple_form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f|
-  = f.error_notification
   .form-inputs
     = f.input :email, required: true, as: :email, autofocus: true
-    - if devise_mapping.confirmable? && resource.pending_reconfirmation?
-      %p= t(".currently_waiting_confirmation_for_email", email: resource.unconfirmed_email)
-    = f.input :password,
-      hint: t(".leave_blank_if_you_don_t_want_to_change_it"),
-      required: false,
-      input_html: { autocomplete: "new-password" }
-    = f.input :password_confirmation,
-      required: false,
-      input_html: { autocomplete: "new-password" }
     = f.input :current_password,
-      hint: t(".we_need_your_current_password_to_confirm_your_changes"),
       required: true,
       input_html: { autocomplete: "current-password" }
+    - if devise_mapping.confirmable? && resource.pending_reconfirmation?
+      %p= t(".currently_waiting_confirmation_for_email", email: resource.unconfirmed_email)
+    #change-password{ class: change_password_fields_class(resource) }
+      = f.input :password,
+        label: t(".new_password"),
+        disabled: password_change_field_disabled?(resource),
+        input_html: { autocomplete: "new-password" }
+      = f.input :password_confirmation,
+        disabled: password_change_field_disabled?(resource),
+        label: t(".new_password_again"),
+        input_html: { autocomplete: "new-password" }
   .form-actions
-    = f.button :submit, t(".update")
+    = f.button :submit, t(".save")
+    %span.options
+      %a#toggle-password{ href: "javascript:;" }
+        = t(".change_password")
+      %span
+        = t(".unhappy")
+        = link_to t(".delete_my_account"), registration_path(resource_name),
+          data: { confirm: t(".are_you_sure") }, method: :delete, class: "warning"
+
 %h3= t(".invite_user")
+= simple_form_for(UserInvitation.new, url: user_invitations_path) do |f|
+  = f.input :recipient_email
+  = f.submit
 
 %table
   %thead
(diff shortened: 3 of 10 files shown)


app/views/users/registrations/edit.html.haml
```diff
@@ -27,19 +42,17 @@
...
+      %td= t(".user_invitations.no_one_yet")
+    - else
+      - @invitations.to_a.each do |invite|
+        %tr{ id: "invite-#{invite.id}" }
+          %td= invite.recipient_email
+          %td= invite.sent_at
+          %td= invite.accepted_at
 
-%h3= t(".cancel_my_account")
+%h3= t(".delete_my_account")
```
ashpil, line 54: "Cancel" felt like somewhat odd terminology. We're not a subscription service.

ashpil:
Also, I think deleting an account should require password confirmation as well, which we don't currently do. This puts us to 3/4ths of things on this page as requiring a password, which makes me think maybe we should just have a password requirement for accessing the page? I suppose this will make more sense to implement when we have a public profile/user editing their own info distinction.

app/views/users/registrations/edit.html.haml
hernanat: agree!

app/views/users/registrations/edit.html.haml
```diff
@@ -1,24 +1,39 @@
 %h2.prompt= t(".title", resource: resource.model_name.human)
+%h3= t(".change_email")
 = simple_form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f|
   = f.error_notification
   .form-inputs
+    = f.input :current_password,
```
hernanat, line 6: we also need the password to change the email?

app/views/users/registrations/edit.html.haml
```diff
@@ -1,24 +1,39 @@
...
+      required: true,
+      input_html: { autocomplete: "current-password" }
     = f.input :email, required: true, as: :email, autofocus: true
     - if devise_mapping.confirmable? && resource.pending_reconfirmation?
       %p= t(".currently_waiting_confirmation_for_email", email: resource.unconfirmed_email)
+  .form-actions
+    = f.button :submit, t(".save")
+
+%h3= t(".change_password")
+= simple_form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f|
```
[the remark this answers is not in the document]
hernanat: That makes sense. The only thing is that I find it a bit awkward having the password field in two places. Maybe we can do something like Reddit or Github where the email / password (and generally, security related things) are in separate views (or a modal -- which we don't currently have a pattern form)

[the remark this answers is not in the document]
ashpil: Yeah, that makes more sense for when we have a larger distinction between a public profile page and an "edit your own info" page.

app/assets/stylesheets/base/_tables.scss
```diff
@@ -1,6 +1,6 @@
 table {
   border-collapse: collapse;
-  margin: $spacing 0;
+  margin-bottom: $spacing 0;
```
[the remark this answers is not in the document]
ashpil: Right, oops

app/views/users/registrations/edit.html.haml
ashpil: Don't we?

app/views/users/registrations/edit.html.haml
```diff
@@ -27,19 +42,17 @@
...
-    - @invitations.to_a.each do |invite|
-      %tr{ id: "invite-#{invite.id}" }
-        %td= invite.recipient_email
-        %td= invite.sent_at
-        %td= invite.accepted_at
-
-= simple_form_for(UserInvitation.new, url: user_invitations_path) do |f|
-  = f.input :recipient_email
-  = f.submit
+    - if !@invitations.nil? && @invitations.empty?
```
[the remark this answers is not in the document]
ashpil: Not sure what you mean here. Just creating an `update` method that's the same as `create` makes the page not load. I just don't know what else it's supposed to be doing.

app/views/users/registrations/edit.html.haml
```diff
@@ -1,24 +1,37 @@
 %h2.prompt= t(".title", resource: resource.model_name.human)
+%h3= t(".change_email")
```
ashpil, line 2: Another issue with this two form system is that simple_form treats them as one. Meaning that if a user inputs a bad password, the validation puts the error in both of the forms, rather than just the inputted one. Still need to fix this.

ashpil:
For some reason on input the url is transformed form `/users/edit` to just `/users` which doesn't seem right to me?

app/views/users/registrations/edit.html.haml
ashpil: Not sure how to go about this.

app/views/users/registrations/edit.html.haml
```diff
@@ -1,24 +1,37 @@
...
+      required: true,
+      input_html: { autocomplete: "current-password" }
     = f.input :email, required: true, as: :email, autofocus: true
     - if devise_mapping.confirmable? && resource.pending_reconfirmation?
       %p= t(".currently_waiting_confirmation_for_email", email: resource.unconfirmed_email)
+  .form-actions
+    = f.button :submit, t(".save")
+
+%h3= t(".change_password")
+= simple_form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f|
```
hernanat, line 15: so really I think that we ought to have two separate forms -- an email edit form and a password edit form, that each get sent to their own controllers. And then we just wouldn't use the default devise registrations edit / update actions. Or we keep the default devise action for editing everything except the password, and then we have a separate form / controller for changing the password. I'll add some changes for that.

ashpil:
I was messing around with my Discord account and saw the way they handle this issue of having two separate forms for password/email and I liked it more than we have here, so I figured I'd implement it.

This is what it looks like by default:
image

And if someone clicks "Change password?"
image

This is a bit cleaner, I think. I also moved the "delete my account" button up a bit, cleaner this way too.

ashpil:
Oh hey, CircleCI is actually complaining about the bundle this time, not failed tests! Yay!

hernanat:
do we need to have the email field on the password form?

app/assets/stylesheets/pages/_users.scss
```diff
@@ -1 +1,26 @@
 @import "users/edit";
+
+
+form {
```
hernanat, line 4: can you change your text editor to use 2 spaces for the scss files of our project to be consistent with our other scss files? :innocent:

app/views/users/registrations/edit.html.haml
```diff
@@ -1,24 +1,36 @@
...
+        label: t(".new_password"),
+        input_html: { autocomplete: "new-password", disabled: "disabled", id: "new_password" }
+      = f.input :password_confirmation,
+        label: t(".new_password_again"),
+        input_html: { autocomplete: "new-password", disabled: "disabled", id: "new_password_again" }
   .form-actions
-    = f.button :submit, t(".update")
+    = f.button :submit, t(".save")
+    %span.options 
+      %a#toggle_password{href: "javascript:;"}
```
hernanat, line 23: what is this doing here?

hernanat:
@ashpil I pushed some changes to fix the linter failures so make sure you pull those down before pushing anything else ~

hernanat:
I'm noticing a UI regression locally -- the "delete my account" section now falls on the same line as "Change password"

hernanat:
> do we need to have the email field on the password form?

ok after playing with this locally I see what's going on. We only display these fields if the user wants to change their password, otherwise we leave them out. I think this makes sense. We can always iterate if anything comes up.

ashpil:
> 
> 
> I'm noticing a UI regression locally -- the "delete my account" section now falls on the same line as "Change password"

Does it look the same as on the screenshot? If so, that was intended.

ashpil:
> 
> 
> do we need to have the email field on the password form?

Ah, that's right. I was copying the Discord model but neglected to realize that the email field is prefilled and thus unchanged if it it untouched. Is this possible to do with simple form?

hernanat:
> > I'm noticing a UI regression locally -- the "delete my account" section now falls on the same line as "Change password"
> 
> Does it look the same as on the screenshot? If so, that was intended.

nope, they're squished right next to each other

app/views/users/registrations/edit.html.haml
ashpil: Is there a better way to do this? I want to get `<a>` tag styling, but it's not an actual link. StackOverflow said just do that, not sure if there's a better way....

app/views/users/registrations/edit.html.haml
```diff
@@ -10,17 +10,17 @@
...
+        input_html: { autocomplete: "new-password", disabled: "disabled", id: "new-password" }
       = f.input :password_confirmation,
         label: t(".new_password_again"),
-        input_html: { autocomplete: "new-password", disabled: "disabled", id: "new_password_again" }
+        input_html: { autocomplete: "new-password", disabled: "disabled", id: "new-password-again" }
   .form-actions
     = f.button :submit, t(".save")
-    %span.options 
-      %a#toggle_password{href: "javascript:;"}
+    %span.option:
```
ashpil: Here was the culprit.

hernanat (approved):

ashpil:
Looks good, thanks!

hernanat:
> Todo:
> 
>     * fix invitations list disappearing on validations

done

PR #65, opened 2020-10-05
Adds footer

ashpil:
This is a simple commit that just adds a footer.

It might look pretty weird until we have some content, though:

image

### DIFF
--- a/app/views/pages/rules.html.haml
+++ b/app/views/pages/rules.html.haml
@@ -1,0 +1,2 @@
+.text
+  Placeholder
--- a/config/locales/navigation.en.yml
+++ b/config/locales/navigation.en.yml
@@ -5,3 +5,8 @@ en:
       submit: "Submit"
       your_submissions: "Your submissions"
       inbox: "Inbox"
+  footer:
+      links:
+          about: "About"
+          rules: "Rules"
+          source: "GitHub"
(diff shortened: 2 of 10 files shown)


app/views/pages/rules.html.haml
```diff
@@ -0,0 +1,2 @@
+.text
+  = "Placeholder"
```
ashpil, line 2: Don't believe we have anything to put here just yet? Are we going to do the whole `config/locales` for large chunks of text like this, too?

config/locales/navigation.en.yml
```diff
@@ -5,3 +5,8 @@ en:
       submit: "Submit"
       your_submissions: "Your submissions"
       inbox: "Inbox"
+  footer:
+      links:
+          about: "About"
+          rules: "Rules"
+          source: "GitHub"
```
ashpil, line 12: Not sure what else we need atm

spec/requests/pages_request_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
ashpil, line 1: The command I used to add a controller just generated this spec page, which I assume is fine? Do we need any tests for the footer?

ashpil: Nvm CircleCI didn't like this so I just chucked it

hernanat (approved):
LGTM after rebase (if needed) + squash

hernanat (approved):

PR #66, opened 2020-10-05
[WIP] reply notifications + fuck ton of tech debt

hernanat:
We want users to be notified when they've received a reply
  to their submissions / comments.

This PR is an ugly-as-fuck way of doing that.

A nicer way would be utilizing a SQL view + maybe separate
  templates for rendering the reply notifications / comments
  instead of trying to cram everything into what we currently
  have.

I'll make a tech debt card / might update things in the next
  few days to be less fugly.


app/views/layouts/_navbar.html.haml
```diff
@@ -18,6 +18,6 @@
             %span= link_to t("navigation.links.your_submissions"),
               user_submissions_path(current_user.username)
           %li
-            %span= link_to t("navigation.links.inbox"), "#"
+            %span= link_to t("navigation.links.inbox"), inbox_path
```
hernanat, line 21: todo: change link color when there are notifications

hernanat:
todo: tests

app/views/comments/_comment.html.haml
```diff
@@ -1,16 +1,17 @@
-- children = @comments_by_parent[comment.id]
+- children = local_assigns[:children] || @comments_by_parent[comment.id]
```
hernanat, line 1: this here (and where it's used elsewhere) in the file is a hacky way of rendering only the comment that is within the context of the notification, and its parent (the comment that it's in reply to) if it's a comment reply.

Honestly this would probably be a lot nicer if we had separate partials for "true" comments and replies.

Voting is basically broken for these and I didn't really feel like getting them to work at present since they're not a hard requirement. I think being able to reply from the inbox is fancy enough for now. Also, I should probably fix the behavior not to autoredirect to the context in the thread, or maybe I shouldn't. I don't know.

app/views/submissions/_submission_list_item.html.haml
```diff
@@ -1,11 +1,15 @@
-.submission.votable{ id: submission.short_id, class: submission_classes_for(submission) }
-  %span.scoring{ id: "scoring_" + submission.short_id }
-    .upvote{ id: "upvote_" + submission.short_id, class: control_classes(submission, :upvote) }
-    .score{ id: "score_" + submission.short_id, class: control_classes(submission, :score) }
-      = submission.score
-    .downvote{ id: "downvote_" + submission.short_id, class: control_classes(submission, :downvote) }
+.submission.votable{ id: submission.short_id,
+  class: submission_classes_for(submission, without_action_classes: local_assigns[:without_action_classes]) }
+  - unless local_assigns[:disable_voting]
```
hernanat, line 3: lol so this is more hacky crap so I don't have to fix a bunch of stuff to get voting working, which basically involves implementing a new sql view which I didn't want to get into today (I might feel different tomorrow)

app/views/submissions/_submission_list_item.html.haml
```diff
@@ -1,11 +1,15 @@
...
+    %span.scoring{ id: "scoring_" + submission.short_id }
+      .upvote{ id: "upvote_" + submission.short_id, class: control_classes(submission, :upvote) }
+      .score{ id: "score_" + submission.short_id, class: control_classes(submission, :score) }
+        = submission.score
+      .downvote{ id: "downvote_" + submission.short_id, class: control_classes(submission, :downvote) }
   .submission-title
     .submission-line1
+      - if local_assigns[:prefix].present?
```
hernanat, line 11: `prefix` is just to prefix the notification with indicator text that tells the user whether the notification is for a comment reply or submission reply.

app/views/submissions/_submission_list_item.html.haml
```diff
@@ -20,9 +24,10 @@
       %span.submission-info
         %span.submission-time
           = submitted_time_text(submission)
-      %span.submission-info
-        = link_to hide_link_text_for(submission), "", class: "hide"
-        = link_to save_link_text_for(submission), "", class: "save"
+      - unless local_assigns[:without_action_classes]
```
hernanat, line 27: actions are also broken without some extra work so I disable them for notifications :)

hernanat:
@ashpil I would pull down the branch and make a second account locally and start talking to yourself in a submission to populate your notifications view.

Here's what it looks like so far:

image

hernanat:
err...that was before I disabled self-reply notifications.

ashpil:
Uhhhh unless I'm understanding the concept of notifications wrong this doesn't seem to work for me? I create a post in one, reply in another, and in the original post there should be notifications in the "Inbox" page? Because that doesn't seem to be working for me now

hernanat:
> Uhhhh unless I'm understanding the concept of notifications wrong this doesn't seem to work for me? I create a post in one, reply in another, and in the original post there should be notifications in the "Inbox" page? Because that doesn't seem to be working for me now

With separate accounts?

ashpil:
> 
> With separate accounts?

Yeah. One in one window, one in a private window. That shouldn't mess with it somehow?

app/controllers/comments_controller.rb
```diff
@@ -7,7 +7,7 @@ def create
 
     if @comment.save
       Vote.upvote.create!(votable: @comment, user: current_user)
-      create_thread_reply_notification(submission, @comment) unless current_user.id == @comment.user_id
+      maybe_create_reply_notification(submission, @comment)
```
ashpil, line 10: I moved the checking logic out of here and into the function. Not sure if this is a good idea/idiomatic at all, but if we checked before calling the function we'd essentially be getting the parent and checking a lot of the same things twice, which didn't make complete sense to me.

hernanat: > Not sure if this is a good idea/idiomatic at all

it is actually an excellent choice, thanks for doing this!

ashpil:
I changed the styles a little bit, it was mostly good already.

image

I think it's the submission/comment dichotomy is clear enough without the prefix, so I removed that. And some other small changes that I think make it easier to read.

Also, now Inbox is highlighted in orange if you have notifications! Not sure just orange is the best choice forever, but it works for now.

app/views/layouts/_navbar.html.haml
```diff
@@ -18,6 +18,9 @@
             %span= link_to t("navigation.links.your_submissions"),
               user_submissions_path(current_user.username)
           %li
-            %span= link_to t("navigation.links.inbox"), inbox_path
+            - if (current_user.thread_reply_notifications.select {|n| !n[:dismissed]} ).empty?
+              %span= link_to t("navigation.links.inbox"), inbox_path
+            - else 
+              %span= link_to t("navigation.links.inbox"), inbox_path, class: "notify"
```
ashpil, line 24: There might be a cleaner way to do this?

ashpil (approved):

hernanat:
[Notifications](https://trello.com/c/rKG9qLtB/50-notifications)

ashpil:
Does this need anything other than the CircleCI stuff?

hernanat:
> Does this need anything other than the CircleCI stuff?

tests

ashpil (approved):
Looks good to me! Nice work!

PR #69, opened 2020-10-12
allows admins / mods to manage tags

hernanat:
https://trello.com/c/JSOQScjN

This commit adds some basic admin functionality
  that allows admins to view, add, and edit tags
  for submissions.

We don't allow doing things like deleting tags.

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -1,0 +1,5 @@
+@charset "utf-8";
+@import "bourbon";
+@import "base/base";
+@import "shared/shared";
+@import "pages/admin/pages";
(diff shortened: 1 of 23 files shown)


hernanat:
[As an admin / moderator I can create a tag](https://trello.com/c/JSOQScjN/18-as-an-admin-moderator-i-can-create-a-tag)

ashpil:
A fully-permissioned user's footer looks like this, now:
image
(which I think is fine, a large footer is a sign of wealth and abundance and that the website is seeking a mate)

ashpil:
Think this is good to go after squash too, @hernanat

ashpil (dismissed):

ashpil (dismissed):

PR #68, opened 2020-10-11
allows user to view saved / hidden submissions

hernanat:
https://trello.com/c/9VhIaIX2

This commit allows users to get to submissions that they've
  hidden / saved via their profile.

PR #70, opened 2020-10-12
allows users to sign out

hernanat:
Not sure why we didn't have this before ... oops.


hernanat:
I put it in the footer ... didn't feel like doing another dropdown...we could always move if we want though?

hernanat:
[Allow users to sign out](https://trello.com/c/AlyrYk0j/53-allow-users-to-sign-out)

ashpil:
It would also actually be good if we got in #68 before this, so we can add this to the row in the footer where logged in user stuff happens

ashpil:
This one good after squash as well I think.

ashpil (dismissed):

ashpil (approved):

hernanat:
thanks @ashpil

PR #71, opened 2020-10-12
submission removals

hernanat:
Admins + mods need to be able to remove submissions that
  break community rules / guidelines.

Examples of removable submissions include things like:

- spam
- off-topic content
- doxxing

More fine-grained permissions as to who can view removed submissions,
  who can view the removal details, etc. will come separately.

### DIFF
--- a/app/search_objects/flattened_submission_search.rb
+++ b/app/search_objects/flattened_submission_search.rb
@@ -10,7 +10,13 @@ class FlattenedSubmissionSearch
   end
 
   def by_short_id(short_id)
-    base_relation(include_hidden: true).friendly.find(short_id)
+    base_relation(include_hidden: true).then { |rel|
+      if user.present? && (user.admin? || user.moderator?)
+        rel
+      else
+        rel.where(removed: false)
+      end
+    }.friendly.find(short_id)
   end
 
   def results_paginator
@@ -29,7 +35,7 @@ class FlattenedSubmissionSearch
         then { |rel| user.present? ? filter_by_action(rel) : rel }.
         then { |rel| filter_by_tag(rel) }.
         then { |rel| filter_by_user(rel) }
-    end
+    end.where(removed: false)
   end
 
   def filter_by_action(rel)
@@ -65,7 +71,7 @@ class FlattenedSubmissionSearch
   end
 
   def base_relation(include_hidden: false)
-    FlattenedSubmission.preload(:tags).then { |rel|
+    FlattenedSubmission.preload(:tags, :submission_removal).then { |rel|
       if user.present?
         with_action_info(rel, include_hidden)
       else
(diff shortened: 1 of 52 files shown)


hernanat (commented):
@ashpil idk if you want to pull this down and play with it and see if it can be designed better.

TODO:

- pagination on removal index page

- (eventually) filters on pagination index page

- tests for un-removing

- make removed submissions + actions on them inaccessible to everyone except admins / mods (right now the model / flag doesn't do anything)


As an aside, this permission stuff is getting to / going to get to the point where we probably want an easy way of defining access + action rules by user type so we don't have to do the same type checking everywhere. I'll probably introduce something like [cancan](https://github.com/ryanb/cancan) or [action_policy](https://github.com/palkan/action_policy) to deal with this.

ashpil:
Also doesn't seem removing things actually removes them yet, right? It seems to have added it to the "removals" page but not actually remove it from the index...

hernanat:
> Also doesn't seem removing things actually removes them yet, right? It seems to have added it to the "removals" page but not actually remove it from the index...

yeah, 4th bullet point under todo :P

ashpil:
The comment view will actually be pretty nice now, because we can prob just integrate it into the comment removals page

ashpil:
>@ashpil idk if you want to pull this down and play with it and see if it can be designed better.

Nah this looks good, I'm not too concerned about the mod view and it's hard to beat a table for stuff like this. I just pushed a small change to make the titles a little less verbose and centering the table cells veritically.

hernanat:
@ashpil I pushed a commit for pagination, restricting saved/hiding + upvoting / downvoting comments and submissions to only those where the submission has not been removed, and also added a bunch of tests.

One thing I was wondering was if this is too restrictive which ... now that I've done it I'm a bit meh about changing it but we can. The behavior that I've implemented is that removed submissions + their comments become completely unavailable, but when we do comment removals we're just going to remove + replace the text. I think I'm fine with this but idk what you think.

app/search_objects/flattened_submission_search.rb
```diff
@@ -10,7 +10,13 @@ def initialize(params, user)
   end
 
   def by_short_id(short_id)
-    base_relation(include_hidden: true).friendly.find(short_id)
+    base_relation(include_hidden: true).then { |rel|
```
hernanat, line 13: this is sort of crude, I'll clean this up when I do a general permissions PR

hernanat:
[Notifications](https://trello.com/c/rKG9qLtB/50-notifications)

hernanat:
[As an admin / moderator I can remove a submission](https://trello.com/c/FHpsAtFj/15-as-an-admin-moderator-i-can-remove-a-submission)

PR #73, opened 2020-10-24
Implements some preliminary auth policies for submissions + removals

hernanat:
https://trello.com/c/Sj9gJyA3

Moving forward we should put all of our logic for determining whether a
  given user is allowed to do something into a policy.

This keeps all of the auth rules in one place, and prevents us from having
  to chase duplicate auth code all over the app when we need to make changes.


app/policies/submission_policy.rb
```diff
@@ -0,0 +1,78 @@
...
+    if record.removed? && removed_for_doxxing?
+      user.present? && admin_or_remover?
+    elsif record.removed?
+      user.present? && author_admin_or_moderator?
+    else
+      true
+    end
+  end
+
+  def edit?
```
hernanat, line 31: we don't currently have editing functionality in place but when we do we'll use these edit / update policies

hernanat (commented):
@ashpil curious your thoughts on these policies. We are extra strict with submissions / removals when a submission has been removed for doxxing. The gist of it is that when a submission has been removed for doxxing, the only people allowed to see it / see the removal record are admins and the person who removed it. My reasoning is basically to reduce the number of unnecessary eyes on the information in the event that we truly have a doxxing situation. I restrict the removals themselves as well because right now in the removal view we include the submission title. which may or may not contain some of the doxxed info.

The general removal information will still be available in the public mod log (when we build it), and I think that for doxxing removals we'll just censor the title and anything else there.

ashpil (approved):

hernanat:
ty @ashpil

hernanat:
[Action policy for cleaner permission strategies](https://trello.com/c/Sj9gJyA3/54-action-policy-for-cleaner-permission-strategies)

Issue #52, opened 2020-07-18
Branding


ashpil:
Now that we've acquired psc.al, I think it's time to talk about the site's logo and name in order to define a more concrete aesthetic that will make the site's design easier in the future. 

## Name

What is this site called? Are we going with "Pascal", "Pscal", "Pasc.al", "Psc.al", or a variation of thereof? 

Lobsters calls themselves "Lobsters", but if I understand correctly, to outgroups it is usually referred to as "lobste.rs", for clarity. Hacker News calls themselves their name without anything like it in the [domain](https://news.ycombinator.com/). 

Given these two examples, I think it would be okay to call the site "Pascal", despite operating under the domain `psc.al`. I'd be inclined to do this. I think this also would look better than the alternatives: 

![image](https://user-images.githubusercontent.com/9152478/87863409-e284b480-c917-11ea-8cd1-78ea2ae65d67.png)
![image](https://user-images.githubusercontent.com/9152478/87863403-d4369880-c917-11ea-8f44-4c9b1c0eab6c.png)
![image](https://user-images.githubusercontent.com/9152478/87863421-06e09100-c918-11ea-8b07-deaf9cdaeebf.png)

I think I'd like this as well if we had the `pasc.al` domain, but right now it just feels somewhat deceiving: 
![image](https://user-images.githubusercontent.com/9152478/87863429-1c55bb00-c918-11ea-9181-affdf9255289.png)

Of course, something like this is also a possibility but I fear it might be confusing/I can't figure out a way to make it look good:
![image](https://user-images.githubusercontent.com/9152478/87863509-c33a5700-c918-11ea-840e-2a3ba739a6e1.png)
![image](https://user-images.githubusercontent.com/9152478/87863596-c5e97c00-c919-11ea-81ca-37958da6b1d1.png)

## Logo

The first thing that comes to my mind given Pascal is Pascal's Triangle, so here's a couple designs based on that idea:

![image](https://user-images.githubusercontent.com/9152478/87863820-fd0d5c80-c91c-11ea-8a4c-0ec5f3c765a7.png)

These are all very rough of course, I put in no more work than is required for a mock-up.

I like the minimalist hexagons on the right. They look the most like an actual logo.

I still have to put more thought into the colors; I just used the default colors the editing tool gave me. 

Let me know your thoughts. As I said, I'm still very unsure about all of these ideas, so I'm just throwing some options out there.

hernanat:
[...]

PR #74, opened 2020-10-24
adds new branding!

ashpil:
Finally adds the branding from #52. This includes logo and name.

One small issue...
image
I didn't think about how tiny the logo would look! Maybe we should switch to the three-hex one

Also, does the text look right?


hernanat (dismissed):
:rocket:

hernanat (approved):
:rocket:

PR #79, opened 2021-01-30
Adds ability for admin / mods to ban users

hernanat:
https://trello.com/c/m3dt7K84


hernanat:
[As an admin / moderator I can ban a user](https://trello.com/c/m3dt7K84/17-as-an-admin-moderator-i-can-ban-a-user)

